PR #2, opened 2024-02-28
CCUBE-1040

shockch4rge:
**Changes**

- migrated folder structure from old repo to new one
- kept only baseline code
- refactored variables to better accustom to the new separate repo
- added base project dependencies

- delete branch

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1040)


src/context-providers/provider.tsx
```diff
@@ -0,0 +1,45 @@
...
+    mode: EFormBuilderMode.ADD_FIELD,
+    view: EFormBuilderView.EDIT_FORM,
+};
+
+// =============================================================================
+// REDUCER
+// =============================================================================
+
+export const mainReducer = (state: TState, action: TAction) => {
+    //
```
keithtxw: Just return the state as a baseline.

src/context-providers/hook.ts
```diff
@@ -0,0 +1,11 @@
+import { useContext } from "react";
+import { MainContext } from "./provider";
+
+export const useMainContext = () => {
+    const { state, dispatch } = useContext(MainContext);
+
+    return {
+        state,
+        dispatch,
```
keithtxw: Not necessary to return `dispatch` since we know for sure we won't be using it like that.

src/context-providers/provider.tsx
```diff
@@ -0,0 +1,45 @@
...
+// =============================================================================
+// HELPER FUNCTIONS
+// =============================================================================
+
+// =============================================================================
+// CONTEXT
+// =============================================================================
+export const MainContext = createContext<TMainContext>({
+    state: DEFAULT_VALUES,
+    dispatch: () => null,
```
keithtxw: we usually explicitly assign it to a noop
```
export const noop = () => {
	// does nothing
};
```

keithtxw (changes requested):

src/context-providers/provider.tsx
```diff
@@ -0,0 +1,46 @@
...
+// =============================================================================
+export const MainContext = createContext<TMainContext>({
+    state: DEFAULT_VALUES,
+    dispatch: noop,
+});
+
+// =============================================================================
+// CONTEXT PROVIDER
+// =============================================================================
+export const MainProvider = ({ children }: ComponentProps<"div">) => {
```
keithtxw:
```suggestion
export const MainProvider = ({ children }: { children: React.ReactNode }) => {
```

src/types.ts
```diff
@@ -0,0 +1,11 @@
+export enum EFormBuilderMode {
+    ADD_FIELD = "add-field",
+    EDIT_FIELD = "edit-field",
+    EDIT_PAGE = "edit-page",
+}
+
+export enum EFormBuilderView {
+    EDIT_FORM = "edit-form",
+    SCHEMA_PREVIEW = "schema-preview",
```
keithtxw: think we wont be having this in the module. It'll be set up by the consumer

keithtxw (dismissed):

src/context-providers/provider.tsx
weili-govtech: this has not been rectified

src/context-providers/provider.tsx
weili-govtech, line 1: understand that we previously commented to do away with the `FormBuilder` prefix but looking at it now, I see the need to move these files into a `builder/` sub-folder and use a `Builder` prefix (e.g. `BuilderContext`, `BuilderProvider`) 

reason is this file will get big as more states are added. we want to split them into different context for better management. in the future, we can have `FieldContext`, `ModalContext`, `PageContext` to handle different aspects of the library.

src/context-providers/hook.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from "react";
+import { MainContext } from "./provider";
+
+export const useMainContext = () => {
```
weili-govtech: can do away with the `Context` suffix, so it can be `useBuilder`

src/context-providers/types.ts
```diff
@@ -0,0 +1,28 @@
...
+// =============================================================================
+// CONTEXT
+// =============================================================================
+
+export type TMainContext = TContextType<TState, TAction>;
+
+export type TContextType<State, Action> = {
+    state: State;
+    dispatch: React.Dispatch<Action>;
+};
```
weili-govtech: type naming can be more specific
- TBuilderState
- TBuilderAction
- TBuilderContextType

src/context-providers/types.ts
```diff
@@ -0,0 +1,28 @@
+import React from "react";
```
weili-govtech: no need to import react

src/schemas/index.ts
weili-govtech, line 1: how are these typings meant to be used?

src/types.ts
weili-govtech: this has not been rectified

src/types.ts
weili-govtech, line 1: can we move these typings into `context-providers/builder/types.ts`

weili-govtech (changes requested):

src/context-providers/types.ts
shockch4rge: eslint complains if react is not imported

src/schemas/index.ts
shockch4rge: will be used for field/page types and validation schemas

weili-govtech (approved):

PR #4, opened 2024-02-29
CCUBE-1041

ShIrInRuTh:
**Changes**
 - Created the side panel
 - Made the side panel collapsible
 - Refactored the code to accommodate future changes

**Additional information**
-   You may refer to this [CCUBE-1041](https://jira.ship.gov.sg/browse/CCUBE-1041)


src/components/panel/base-panel-header.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const BasePanelHeader = () => {
+    const { state, togglePanel } = useBuilder();
+    return (
+        <BasePanelHeaderWrapper>
+            <IconButton
+                $iconSize="1.5rem"
+                $iconColor={Color.Neutral[3]}
+                onClick={() => togglePanel(!state.showPanel)}
+            >
+                {state.showPanel ? <ChevronLeftIcon /> : <ChevronRightIcon />}
```
keithtxw: a nice touch would be to rotate it as it change. But i'll accept it as this for now.

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,44 @@
...
+    transform: ${({ $isCollapsed }) => $isCollapsed && "translateX(85%)"};
+    transition: ${Transition.Base};
+    background: ${Color.Neutral[8]};
+    border-left: 1px solid ${Color.Neutral[5]};
+    position: fixed;
+    top: 0;
+    right: 0;
+    box-shadow: 0px 2px 12px 0px rgba(104, 104, 104, 0.25);
+
+	${MediaQuery.MaxWidth.desktopM} {
```
keithtxw: Indentation looks off. Could you help check?

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,44 @@
+import { Color } from "@lifesg/react-design-system/color";
+import { MediaQuery } from "@lifesg/react-design-system/media";
+import { Transition } from "@lifesg/react-design-system/transition";
+import styled from "styled-components";
+
+export const BasePanelWrapper = styled.div<{ $isCollapsed: boolean }>`
```
keithtxw: introduce a styled interface instead
```suggestion
export const Wrapper = styled.div<IWrapperStyleProps>`
```

keithtxw: Can just call this `Wrapper`

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,44 @@
...
+    top: 0;
+    right: 0;
+    box-shadow: 0px 2px 12px 0px rgba(104, 104, 104, 0.25);
+
+	${MediaQuery.MaxWidth.desktopM} {
+		width: 40vw;
+	}    
+`;
+
+export const BasePanelHeaderWrapper = styled.div`
```
keithtxw: Similar to the top comment, can omit the `BasePanel` prefix. Likewise for the elements below

src/components/panel/base-panel.tsx
```diff
@@ -0,0 +1,28 @@
+import { Layout } from "@lifesg/react-design-system/layout";
+import { ComponentProps } from "react";
+import { useBuilder } from "../../context-providers/builder/hook";
+import { BasePanelHeader } from "./base-panel-header";
+import { BasePanelWrapper } from "./base-panel.styles";
+import { Toolbar } from "./toolbar";
+
+export interface IBasePanelProps extends ComponentProps<"div"> {
```
keithtxw: The practice is to have the interfaces specified in a `types.ts` file that sits in `panel` folder. Refer to CONTRIBUTING.md for more info. 

And also don't need to extend `ComponentProps`. Just specify children as a prop
```
children: React.ReactNode;
```

src/components/panel/toolbar/toolbar.styles.tsx
```diff
@@ -0,0 +1,42 @@
...
+    align-items: center;
+    padding: 2rem 1rem;
+    border: 1px solid ${Color.Neutral[5]};
+    border-style: none solid none none;
+    gap: 2rem;
+`;
+
+export const ModeButton = styled.button<{ $active: boolean }>`
+    background: ${({ $active }) =>
+        $active ? Color.Accent.Light[4] : "transparent"};
```
keithtxw: you will need to pass the `props` to the colour, or else the theming won't work. 

Suggest to do as such:
```
background: ${(props) => props.$active ? Color.Accent.Light[4](props) : "transparent"};
```

src/components/template-page/template-page.tsx
keithtxw, line 1: Slightly curious what does this component do? I think it should not have a page per say. It should take up 100% width and height based on it's parent container. 

Let the page setting be dealt with by the web app using this component

src/context-providers/builder/types.ts
```diff
@@ -3,20 +3,26 @@ import { TContextType } from "../types";
...
 export type TBuilderState = {
     mode: EFormBuilderMode;
+    showPanel: boolean;
 };
 
 // =============================================================================
 // ACTIONS
 // =============================================================================
-export type TBuilderAction = Record<string, never>;
+export type TTogglePanel = {
```
keithtxw:
```suggestion
export type TTogglePanelAction = {
```

src/context-providers/builder/types.ts
```diff
@@ -3,20 +3,26 @@ import { TContextType } from "../types";
...
-    EDIT_PAGE = "edit-page",
+    EDIT_PAGES = "edit-page",
 }
 
 // =============================================================================
 // STATE
 // =============================================================================
 export type TBuilderState = {
     mode: EFormBuilderMode;
+    showPanel: boolean;
```
keithtxw: perhaps stating `showSidePanel` might be clearer

src/components/panel/base-panel.tsx
keithtxw, line 1: Would think it is clearer if we name this SidePanel

src/components/common/icon-button/icon-button.styles.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface IconButtonStyleProps {
+    $iconSize?: string;
+    $iconColor?: (props: any) => string;
+}
+
+export const IconButton = styled.button<IconButtonStyleProps>`
+    padding: 0;
+    border: none;
+    display: grid;
```
keithtxw: this style and the one below does nothing since there is no height or width specified. In other words, the button is taking the size of the svg. There is no centering required

src/components/template-page/template-page.tsx
keithtxw: What i think u will need is a main wrapper. A base content panel and a side panel. With the main wrapper taking 100% width and height. Then in the stories folder, can create a file called `doc-elements.tsx` that has a container that takes 100vh and 100vw. Then have `FormBuilder` sit in it.

keithtxw (changes requested):

src/components/panel/base-panel.styles.tsx
keithtxw: i realised the indentation for all the files seem off. Can help to check. Make sure formatting is set to auto upon save.

keithtxw: I checked in code and it looks ok. Maybe just the way Github prints it.

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { IHeaderStyleProps, IWrapperStyleProps } from "./types";
+
+export const Wrapper = styled.div<IWrapperStyleProps>`
+    height: 100%;
+    width: 36.9rem;
+    transform: ${({ $isCollapsed }) => $isCollapsed && "translateX(85%)"};
+    transition: ${Transition.Base};
+    background: ${Color.Neutral[8]};
+    border-left: 1px solid ${Color.Neutral[5]};
+    position: fixed;
```
weili-govtech: as the header and title bar are separated from this library, we are not able to accommodate their heights here. with that, we  should switch from fixed positioning to sticky positioning to keep the panel at the same position when there is scrolling.

at the same time, in order for this panel to stick to the top on scrolling, we will need to let the FormBuilder wrapper instead of the body do the overflow. so what needs to be done:
- update FormBuilder to have 100% height and overflow-y: auto
- update panel to use sticky positioning with top: 0

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { MediaQuery } from "@lifesg/react-design-system/media";
+import { Transition } from "@lifesg/react-design-system/transition";
+import { ChevronLeftIcon } from "@lifesg/react-icons/chevron-left";
+import styled from "styled-components";
+import { IHeaderStyleProps, IWrapperStyleProps } from "./types";
+
+export const Wrapper = styled.div<IWrapperStyleProps>`
+    height: 100%;
+    width: 36.9rem;
+    transform: ${({ $isCollapsed }) => $isCollapsed && "translateX(85%)"};
```
weili-govtech: instead of an arbitrary 85% translate, we can use `calc()` to be more accurate and for better readability, it should be `calc(100% - <width of toolbar>);`

src/components/panel/base-panel.tsx
```diff
@@ -0,0 +1,23 @@
...
+    const { state } = useBuilder();
+    return (
+        <Wrapper $isCollapsed={state.showPanel}>
+            {!disableToolbar && (
+                <>
+                    <BasePanelHeader />
+                    <Toolbar />
+                </>
+            )}
+            {!state.showPanel && children}
```
weili-govtech: will you be fixing the bug to ensure the toolbar and children are in the same row?

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,53 @@
...
+export const HeaderWrapper = styled.div`
+    grid-column: 2 / span 1;
+    max-height: 5.75rem;
+    display: flex;
+    align-items: center;
+    width: auto;
+    gap: 2rem;
+    padding: 2rem;
+    border-bottom: 1px solid ${Color.Neutral[5]};
+`;
```
weili-govtech: - `grid-column` is not needed as it is not applicable for flexbox
- `max-height` is not needed as it not doing anything
- `width` is not needed as it is already auto by default

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,53 @@
...
+    flex-direction: column;
+    width: 100%;
+    overflow-y: scroll;
+`;
+
+export const HeaderIcon = styled(ChevronLeftIcon)<IHeaderStyleProps>`
+    transform-origin: bottom-left;
+    transform: ${({ $isCollapsed }) =>
+        $isCollapsed ? "rotate(-180deg)" : "rotate(0)"};
+    transition: transform 0.1s ease-in-out;
```
weili-govtech: - remove `transform-origin` as the value is not valid, current center origin works well already
- why use 0.1s for the transition when the panel translation animation takes 0.3s?

src/components/panel/base-panel.styles.tsx
```diff
@@ -0,0 +1,53 @@
...
+    padding: 2rem 2rem 0 0;
+    flex: 1;
+`;
+
+export const BasePanelContent = styled.div`
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    overflow-y: scroll;
+`;
```
weili-govtech: not sure how they are meant to be used, can we remove them since they are not used yet?

src/components/panel/toolbar/toolbar.styles.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const ModeButtonList = styled.div`
+    display: flex;
+    flex-direction: column;
+    flex: 1;
+    align-items: center;
+    padding: 2rem 1rem;
+    border: 1px solid ${Color.Neutral[5]};
+    border-style: none solid none none;
+    gap: 2rem;
+`;
```
weili-govtech: why the need to have 2 components, can we combine them?

src/components/panel/base-panel-header.tsx
```diff
@@ -0,0 +1,21 @@
...
+    return (
+        <HeaderWrapper>
+            <IconButton
+                $iconSize="1.5rem"
+                $iconColor={Color.Neutral[3]}
+                onClick={() => togglePanel(!state.showPanel)}
+            >
+                <HeaderIcon $isCollapsed={state.showPanel} />
+            </IconButton>
+            <Text.H3 weight="semibold">Add elements</Text.H3>
```
weili-govtech: this should be rendered from a prop

src/components/panel/base-panel-header.tsx
```diff
@@ -0,0 +1,21 @@
...
+        </HeaderWrapper>
```
weili-govtech: - add a "Save Changes" button
- add a `onClickSaveChanges` prop
- show button if `onClickSaveChanges` is defined

src/context-providers/builder/provider.tsx
```diff
@@ -1,18 +1,19 @@
+import { noop } from "lodash";
```
weili-govtech: use sub-directory import to reduce bundle size

```suggestion
import noop from "lodash/noop";
```

src/main.tsx
```diff
@@ -1,5 +1,20 @@
+import { BasePanel } from "./components/panel/side-panel";
+import {
+    BasePanelBody,
+    BasePanelContent,
+} from "./components/panel/side-panel.styles";
```
weili-govtech: components can be barrel exported in `panel/` and `components/` folders

actually there should be another component to manage which panel to display (elements, pages, details) and that should hold `BasePanelBody` and `BasePanelContent`.however to avoid scope creep, you can move these components into that component and skip the logic to decide which panel to display.

src/components/panel/side-panel.tsx
```diff
@@ -0,0 +1,34 @@
+import { useBuilder } from "../../context-providers/builder/hook";
+import { BasePanelHeader } from "./side-panel-header";
+import { ContentWrapper, Wrapper } from "./side-panel.styles";
+import { Toolbar } from "./toolbar";
+import { IBasePanelProps } from "./types";
+
+export const BasePanel = ({
```
weili-govtech: since the file has been renamed as `side-panel`, can rename this component as `SidePanel`

src/context-providers/builder/provider.tsx
```diff
@@ -48,8 +52,10 @@ export const BuilderProvider = ({
...
 
     return (
-        <BuilderContext.Provider value={{ state, dispatch }}>
-            {children}
-        </BuilderContext.Provider>
+        <MainWrapper>
+            <BuilderContext.Provider value={{ state, dispatch }}>
+                {children}
+            </BuilderContext.Provider>
+        </MainWrapper>
```
weili-govtech: provider file should only render the necessary components for this context provider and not render ui elements, move `MainWrapper` out of this component

weili-govtech (changes requested):

src/components/panel/base-panel.tsx
ShIrInRuTh: yep this has been fixed in the commit I made yesterday morning

src/components/panel/base-panel.styles.tsx
ShIrInRuTh: This is already being used in the commit I made yesterday morning and is used as styling for the content in the side panel. This is a temporary implementation; later on for 1042 i will be moving this into a separate folder.

`export const FormBuilder = () => {
    return (
        <BuilderProvider>
            <SidePanel>
                <SidePanelBody>
                    <SidePanelContent>
                        <h2>Hello World!</h2>
                    </SidePanelContent>
                </SidePanelBody>
            </SidePanel>
        </BuilderProvider>
    );
};`

src/components/panel/toolbar/toolbar.styles.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const ToolbarWrapper = styled.div`
+    height: 100%;
+    background: ${Color.Neutral[8]};
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    width: fit-content;
+    padding: 2rem 1rem;
+    border: 1px solid ${Color.Neutral[5]};
+    border-style: none solid none none;
```
weili-govtech:
```suggestion
    border-right: 1px solid ${Color.Neutral[5]};
```

src/components/panel/base-panel-header.tsx
weili-govtech: this is not fixed yet?

src/components/panel/base-panel-header.tsx
weili-govtech: thanks for adding the button, however I don't see the `onClickSaveChanges` prop

src/components/panel/panel.tsx
```diff
@@ -0,0 +1,17 @@
...
+} from "./side-panel/side-panel.styles";
+
+export const Panel = () => {
+    return (
+        <SidePanel>
+            <SidePanelBody>
+                <SidePanelContent>
+                    <h2>Hello World!</h2>
+                </SidePanelContent>
+            </SidePanelBody>
```
weili-govtech: - SidePanelBody should be within the SidePanel component
- move `overflow-y` style from SidePanelContent to SidePanelBody
- understand SidePanelContent is likely just a placeholder component, I don't see much use of it at this point and think we can remove it altogether, just render `<h2>Hello World!</h2>` will do

weili-govtech (changes requested):

src/components/panel/panel.tsx
keithtxw: We've refactored the layout to make things clearer

keithtxw (dismissed):

src/components/common/icon-button/icon-button.styles.tsx
```diff
@@ -0,0 +1,30 @@
+import styled from "styled-components";
+
+// =============================================================================
+// STYLING INTERFACE
+// =============================================================================
+interface IconButtonStyleProps {
```
weili-govtech: prefix with `I` for interfaces

```suggestion
interface IIconButtonStyleProps {
```

src/components/side-panel/side-panel.styles.tsx
```diff
@@ -0,0 +1,49 @@
...
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    width: 36.8rem;
+    transition: ${Transition.Base};
+    background: ${Color.Neutral[8]};
+    border-left: 1px solid ${Color.Neutral[5]};
+    margin-left: auto;
+    box-shadow: 0px 2px 12px 0px rgba(104, 104, 104, 0.25);
+    position: fixed;
```
weili-govtech, line 25: as the header and title bar are separated from this library, we are not able to accommodate their heights here. with that, we should switch from fixed positioning to sticky positioning to keep the panel at the same position when there is scrolling.

weili-govtech (changes requested):

src/components/side-panel/side-panel.styles.tsx
weili-govtech: my apologies, it appears that the sticky positioning is not working well, can you revert to fixed positioning first? we will come up with an alternate solution to this in another ticket.

weili-govtech (approved):

PR #7, opened 2024-03-06
CCUBE-1061

shockch4rge:
**Changes**
- base card component for listing UI
- active/inactive state represented
- no functionality in buttons

-  delete branch

**Additional information**

-   You may refer to this [CCUBE-1061](https://jira.ship.gov.sg/browse/CCUBE-1061)
-   Will push storybook variant in the future


src/components/card/list/list.tsx
keithtxw, line 1: I think this should be in a separate folder of its own instead of within `card`

src/components/card/card.tsx
```diff
@@ -0,0 +1,63 @@
...
+} from "./card.styles";
+import { BinIcon } from "@lifesg/react-icons/bin";
+import { CopyIcon } from "@lifesg/react-icons/copy";
+import { EnvelopeIcon } from "@lifesg/react-icons/envelope";
+import { EFieldType } from "src/schemas/types";
+import { Text } from "@lifesg/react-design-system/text";
+
+export const Card = ({ field }: ICardProps) => {
+    const { getFocusControls } = useBuilder();
+    const { isFocused, setFocus } = getFocusControls(field.internalId);
```
keithtxw: I'd like this to be coming in as a prop instead as I hope to reuse the component in the element listing in the side panel

src/context-providers/builder/provider.tsx
```diff
@@ -4,15 +4,31 @@ import {
...
+import { TField } from "src/schemas";
+import { EFieldType } from "src/schemas/types";
 
 // =============================================================================
 // DEFAULT VALUES
 // =============================================================================
 const DEFAULT_VALUES: TBuilderState = {
-    mode: EFormBuilderMode.ADD_FIELD,
+    mode: EBuilderMode.ADD_FIELD,
+    fields: new Map<string, TField>()
```
keithtxw: able to remove this since the initial state is an empty display. To test the card's look, create another story

keithtxw (changes requested):

src/components/common/spacer/spacer.styles.tsx
```diff
@@ -0,0 +1,12 @@
+import styled, { css } from "styled-components";
+
+export const Spacer = styled.div<{ $width?: string }>`
```
keithtxw: what's this for? How does it differ from adding styles to `CardBody`?

src/schemas/base-field.ts
```diff
@@ -0,0 +1,8 @@
+import { EFieldType } from "./types";
+
+export type TBaseField = {
```
keithtxw: change to `interface`

src/schemas/index.ts
```diff
@@ -0,0 +1,3 @@
+import { TBaseField } from "./base-field";
+
+export type TField = TBaseField;
```
keithtxw: Don't see a need to assign another name to it. We'll extend the `BaseField` interface if required

keithtxw (changes requested):

src/schemas/index.ts
shockch4rge: forgot to mention that `TBaseField` is a placeholder for a type union of future fields, basically it will look something like `TField = TEmailField | TNumericField`

src/context-providers/builder/hook.ts
```diff
@@ -1,10 +1,26 @@
 import { useContext } from "react";
 import { BuilderContext } from "./provider";
+import { TFocusFieldAction } from "./types";
 
 export const useBuilder = () => {
-    const { state } = useContext(BuilderContext);
+    const { state, dispatch } = useContext(BuilderContext);
+
+    const getFocusControls = (internalId: string) => {
```
keithtxw: wrap this in a useCallback

keithtxw (dismissed):

src/components/card/card.styles.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const CardContent = styled.div<ICardContentStyles>`
+    padding: 1rem;
+    border-radius: 0.5rem;
+    position: relative;
+    display: flex;
+    align-items: center;
+    box-shadow: 0px 2px 8px 0px rgba(104, 104, 104, 0.24);
+
+    :hover {
+        border: 1px solid ${Color.Primary};
```
weili-govtech: adding border on hover is causing the contents to jump on hover, you should add a transparent 1px border by default to prevent this

src/components/card/card.styles.tsx
```diff
@@ -0,0 +1,91 @@
+import styled, { css } from "styled-components";
+import { ICardContentStyles } from "./types";
+import { DragHandleIcon } from "@lifesg/react-icons/drag-handle";
+import { Button } from "@lifesg/react-design-system/button";
+import { Color } from "@lifesg/react-design-system/color";
+import { Spacer } from "../common/spacer";
+
+export const CardContent = styled.div<ICardContentStyles>`
```
weili-govtech: for the different states, I notice it differs from [figma](https://www.figma.com/file/G9h7ux7jd3IGnJn6OEfDhC/%5BWeb%5D-Sequential?type=design&node-id=3903-50805&mode=design&t=KatRg30j2NGo692p-0), e.g. on hover, the box shadow is in primary colour, the dragger is missing, duplicate and delete colours are different from the figma

stories/card/card.stories.tsx
weili-govtech, line 1: can we have a story for the card rendered in 2 columns?

src/components/card/card.tsx
```diff
@@ -0,0 +1,60 @@
+import { useBuilder } from "src/context-providers/builder";
```
weili-govtech: remove as it is not used

weili-govtech (changes requested):

src/components/list/list.tsx
weili-govtech, line 1: this component is not needed will likely have a conflict with the listing panel implementation
if there is a need for a separate wrapper to render the cards in storybook, the component should exist in storybook only

`desktopCols={true ? [3, 11] : [1, 7]}` is also redundant because it is always truthy

src/components/card/card.tsx
```diff
@@ -0,0 +1,84 @@
...
+    onClick,
+    onHover,
+    ...colProps
+}: ICardProps) => {
+    const [isHovered, setIsHovered] = useState(false);
+
+    const spansFull =
+        !colProps.desktopCols ||
+        (Array.isArray(colProps.desktopCols) &&
+            colProps.desktopCols.reduce((a, b) => a + b, 0) === 14);
```
weili-govtech: instead of trying to determine the width based on desktopCols, the card should component should have a prop to define whether it takes full or half width, with that, you can then assign the `desktopCols` accordingly

src/components/card/types.ts
```diff
@@ -0,0 +1,13 @@
+import { ColProps } from "@lifesg/react-design-system";
+import { TField } from "src/schemas";
+
+export interface ICardProps extends ColProps {
+    field: TField;
+    isFocused?: boolean;
+    onHover?: () => void;
```
weili-govtech: what is the hover callback meant for?

src/components/common/spacer/spacer.styles.tsx
weili-govtech: @shockch4rge able to address this?

src/components/card/card.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <CardContent
+            {...colProps}
+            $isFocused={isFocused}
+            onClick={onClick}
+            onMouseEnter={handleHover}
+            onMouseLeave={() => setIsHovered(false)}
+        >
+            <CardBody>
+                {/* TODO: placeholder dragger icon */}
+                {(isFocused || isHovered) && <CardDragger />}
```
weili-govtech: can we have a simple animation of the icon and text moving to the right when the dragger appears?

src/context-providers/builder/provider.tsx
```diff
@@ -1,18 +1,20 @@
...
 import { createContext } from "react";
 import { useImmerReducer } from "use-immer";
 import {
-    EFormBuilderMode,
+    EBuilderMode,
     TBuilderAction,
     TBuilderContext,
     TBuilderState,
 } from "./types";
+import { noop } from "lodash";
```
weili-govtech, line 9: should organise import

src/context-providers/builder/types.ts
```diff
@@ -13,8 +11,10 @@ export enum EFormBuilderMode {
 // STATE
 // =============================================================================
 export type TBuilderState = {
-    mode: EFormBuilderMode;
+    mode: EBuilderMode;
     showSidePanel: boolean;
+    fields: Map<string, TField>;
+    focusedField: TField | null;
```
weili-govtech: instead of `null`, can we update `focusedField` to be optional, this way it can accept `undefined` 

```suggestion
    focusedField?: TField;
```

src/context-providers/builder/hook.ts
```diff
@@ -11,10 +12,25 @@ export const useBuilder = () => {
         });
     };
 
+    const getFocusControls = useCallback((internalId: string) => {
+        const setFocus = (payload: TFocusFieldAction["payload"]) => {
```
weili-govtech:
```suggestion
        const setFocus = (payload?: TField) => {
```

src/context-providers/builder/types.ts
```diff
@@ -25,7 +25,12 @@ export type TTogglePanelAction = {
     payload: boolean;
 };
 
-export type TBuilderAction = TTogglePanelAction;
+export type TBuilderAction = TTogglePanelAction | TFocusFieldAction;
+
+export type TFocusFieldAction = {
+    type: "focus-field";
+    payload: TField | null;
```
weili-govtech:
```suggestion
    payload?: TField;
```

src/context-providers/builder/provider.tsx
```diff
@@ -1,18 +1,20 @@
...
 
 // =============================================================================
 // DEFAULT VALUES
 // =============================================================================
 const DEFAULT_VALUES: TBuilderState = {
-    mode: EFormBuilderMode.ADD_FIELD,
+    mode: EBuilderMode.ADD_FIELD,
+    fields: new Map(),
+    focusedField: null,
```
weili-govtech:
```suggestion
    focusedField: undefined,
```

src/schemas/base-field.ts
```diff
@@ -0,0 +1,8 @@
+import { EFieldType } from "./types";
+
+export type IBaseField = {
+    id: string;
+    internalId: string;
+    label: string;
+    type: EFieldType;
+};
```
weili-govtech: this should be an interface right? other field types can then extend from `IBaseField`

weili-govtech (changes requested):

src/context-providers/builder/hook.ts
shockch4rge: i think it would be better if we had one source of truth for the action payload's type

src/components/card/card.styles.tsx
keithtxw: Due to the blockage for other parts, we will do the dragger in another ticket

keithtxw:
Rewrote many components. Requesting for a re-review

src/components/element-card/element-card.tsx
```diff
@@ -0,0 +1,101 @@
...
+                </DetailsContainer>
+                {isFocused && (
+                    <ActionsContainer>
+                        <ActionButton
+                            type="button"
+                            onClick={handleDuplicateClick}
+                        >
+                            <CopyIcon />
+                            Duplicate
+                        </ActionButton>
```
weili-govtech: - should have `$primary` by default
- need a disabled state

src/components/common/base-card/base-card.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+interface IProps {
+    children: React.ReactNode;
+    focused?: boolean;
+    onClick?: (event: React.MouseEvent<HTMLDivElement>) => void;
+}
+
+export const BaseCard = ({ children, focused, onClick }: IProps) => {
+    return (
+        <Body tabIndex={0} $focused={focused} onClick={onClick}>
```
weili-govtech, line 11: rename as `CardBody` to be clearer

src/components/common/base-card/base-card.styles.tsx
```diff
@@ -0,0 +1,37 @@
...
+                  border: 1px solid ${Color.Primary};
+                  background: ${Color.Accent.Light[5]};
+              `
+            : css`
+                  border: 1px solid transparent;
+              `}
+
+    :hover {
+        border-color: ${Color.Primary};
+    }
```
weili-govtech, line 37: understand we are deviating from figma by not applying a different box-shadow colour, can we have a TODO comment to indicate that? this affects both hover and hover in focused states

weili-govtech (changes requested):

src/components/common/base-card/base-card.tsx
keithtxw: But it is within the BaseCard though. I thought it would be quite clear

src/components/element-card/element-card.tsx
keithtxw: From the looks of it, the duplicate was always greyed out. As for the disabled state, I can't find it in the Figma. Do u have a link?

weili-govtech: ok you can refer to this: https://www.figma.com/file/G9h7ux7jd3IGnJn6OEfDhC/%5BWeb%5D-Sequential?type=design&node-id=2963-22032&mode=design&t=OpLVEogw3w6KO5nU-0

src/components/common/base-card/base-card.tsx
weili-govtech: I actually mistook it as the `<body>` initially but if it does not look that misleading, I am ok to close it

keithtxw: I am referencing the style file so i think it is fine. It's an internal component anyway

weili-govtech (approved):

PR #22, opened 2024-03-28
CCUBE-1062

ShIrInRuTh:
**Changes**

1. Ammended handleClick function for element card to work even when there is no elements in focused state
2. Add a new context to cater to the toggling of the edit element panel

**Additional information**

-   You may refer to this [CCUBE-1062](https://jira.ship.gov.sg/browse/CCUBE-1062) for more info about the ticket

### DIFF
--- a/src/components/main-panel/main-panel.tsx
+++ b/src/components/main-panel/main-panel.tsx
@@ -1,6 +1,5 @@
 import { ErrorDisplay } from "@lifesg/react-design-system/error-display";
-import { EBuilderMode, useBuilder } from "src/context-providers";
-import { TElement } from "src/schemas";
+import { TElement, useBuilder } from "src/context-providers";
 import { ElementCard } from "../element-card";
 import {
     ElementItemWrapper,
--- a/src/components/main-panel/main-panel.tsx
+++ b/src/components/main-panel/main-panel.tsx
@@ -1,5 +1,5 @@
 import { ErrorDisplay } from "@lifesg/react-design-system/error-display";
-import { useBuilder } from "src/context-providers";
+import { EBuilderMode, useBuilder } from "src/context-providers";
 import { TElement } from "src/schemas";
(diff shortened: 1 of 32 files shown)


src/context-providers/builder/provider.tsx
```diff
@@ -59,6 +59,10 @@ export const builderReducer = (
             state.focusedElement = null;
             break;
         }
+        case "toggle-edit-panel": {
+            state.focusedElement = action.payload;
+            break;
+        }
```
weili-govtech: is this action type intended? I don't see it being used

src/components/main-panel/main-panel.tsx
```diff
@@ -21,20 +21,22 @@ export const MainPanel = () => {
...
     const handleElementCardClick = (element: TElement) => () => {
-        if (focusedElement.element.internalId !== element.internalId) {
+        if (focusedElement?.element?.internalId !== element.internalId) {
             /**
              * TODO: Add check if element is dirty when setting focus
              * Hardcode to false for now
              */
             focusElement(element, false);
         }
+        toggleEditPanel(element, true);
```
weili-govtech: `toggleEditPanel` also dispatches a `focus-element` action and in this case, the `isDirty` is set to true which is the opposite of what line 37 is doing.

instead of firing twice with `toggleEditPanel`, you should use `togglePanel`

with this change, the `toggleEditPanel` action is not needed at all and can be removed

weili-govtech: I also don't think the focusedElement check is needed, can we remove it?

weili-govtech (changes requested):

weili-govtech:
thanks for the updates, do resolve the merge conflicts before we approve the PR

keithtxw (approved):

weili-govtech (approved):

PR #25, opened 2024-04-02
CCUBE-1202

ShIrInRuTh:
**Changes**

-   Ensure that when the close button is clicked, it will display the panel the user last accessed
- If the user clicked on the field to display the edit element panel, upon pressing on the close button it should cause the panel to be collapsed.

**Additional information**

-   You may refer to this [CCUBE-1202](https://jira.ship.gov.sg/browse/CCUBE-1202)
-   Will need to re-test once [CCUBE-1062](https://jira.ship.gov.sg/browse/CCUBE-1062) is pushed to main, as for now the element card is not clickable, so I am not able to test the other use case scenarios

### DIFF
--- a/src/components/side-panel/side-panel-header.tsx
+++ b/src/components/side-panel/side-panel-header.tsx
@@ -18,7 +18,7 @@ export const SidePanelHeader = () => {
         togglePanel,
         toggleMode,
         removeFocusedElement,
-        pastMode,
+        focusedElement,
     } = useBuilder();
 
     // =========================================================================
@@ -31,10 +31,9 @@ export const SidePanelHeader = () => {
     //     return "Save button clicked";
     // };
     const handleCrossButtonClick = () => {
-        toggleMode(pastMode.panelMode);
-        togglePanel(pastMode.panelState);
         removeFocusedElement();
-
+        toggleMode(currentMode);
+        togglePanel(showSidePanel);
         // TODO: Run some validation to check if the element is valid. If not, delete the element
     };
 
@@ -42,11 +41,13 @@ export const SidePanelHeader = () => {
     // HELPER FUNCTIONS
     // =========================================================================
     const getHeaderTitle = () => {
+        if (focusedElement) {
+            return "Edit details";
+        }
+
         switch (currentMode) {
             case EBuilderMode.ADD_ELEMENT:
                 return "Add elements";
-            case EBuilderMode.EDIT_ELEMENT:
-                return "Edit details";
             case EBuilderMode.EDIT_PAGES:
                 return "Add pages";
         }
@@ -55,6 +56,18 @@ export const SidePanelHeader = () => {
     // RENDER FUNCTIONS
     // =========================================================================
     const renderIconButton = () => {
+        if (focusedElement) {
+            return (
+                <IconButton
+                    $iconSize="1.5rem"
+                    $iconColor={Color.Neutral[3]}
+                    onClick={handleCrossButtonClick}
+                >
+                    <CrossIcon />
+                </IconButton>
+            );
+        }
+
         switch (currentMode) {
             case EBuilderMode.ADD_ELEMENT:
             case EBuilderMode.EDIT_PAGES:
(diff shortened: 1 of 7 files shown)


src/context-providers/builder/provider.tsx
```diff
@@ -17,6 +17,7 @@ const DEFAULT_VALUES: IBuilderState = {
     focusedElement: null,
     showSidePanel: true,
     orderedIdentifiers: [],
+    pastMode: {},
```
weili-govtech: as the provider can store various states not limited to the side panel, we will need to be more specific with the naming, suggest to use `pastSidePanelState`

src/context-providers/builder/hook.ts
```diff
@@ -6,13 +6,15 @@ import { EBuilderMode, IElementIdentifier } from "./types";
...
-        });
-    }, []);
+    const togglePanel = useCallback(
+        (isCollapsed: boolean) => {
+            dispatch({
+                type: "toggle-panel",
+                payload: isCollapsed,
+            });
+        },
+        [state.mode, state.showSidePanel]
```
weili-govtech: why the need for these 2 dependencies?

src/context-providers/builder/hook.ts
```diff
@@ -82,35 +91,49 @@ export const useBuilder = () => {
...
-        });
-    }, []);
+    const toggleMode = useCallback(
+        (mode: EBuilderMode) => {
+            dispatch({
+                type: "toggle-mode",
+                payload: mode,
+            });
+        },
+        [state.mode, state.showSidePanel]
```
weili-govtech: dependencies are not needed

weili-govtech (changes requested):

src/context-providers/builder/provider.tsx
keithtxw: I do not think it is a good idea to keep reference to the past state. Maybe we should decouple showing the edit panel as a `mode`. Instead we can use the `focusedElement` as the reference. If it is not null, it means we will show the edit element panel.

Likewise for the side panel, if `focusedElement`is not null, show the side panel. Then the original `showSidePanel` should still be holding the value prior to focusing the element.

src/context-providers/builder/hook.ts
```diff
@@ -24,40 +23,37 @@ export const useBuilder = () => {
...
-                        element: newElement,
-                        isDirty: true,
-                    },
-                });
-            }
-        },
-        [state.orderedIdentifiers]
-    );
+        }
+    }, []);
```
weili-govtech: with `useCallback`, the entire function will be memo-ised aka cached. this means that variables used within it will not be updated, hence if there are external non-argument variables used within it, they need to be added as dependencies. in this case there is a need to add `state.orderedIdentifiers`

src/context-providers/builder/hook.ts
```diff
@@ -24,40 +23,37 @@ export const useBuilder = () => {
         [state.orderedIdentifiers]
```
weili-govtech: not needed

src/context-providers/builder/provider.tsx
```diff
@@ -17,6 +17,7 @@ const DEFAULT_VALUES: IBuilderState = {
     focusedElement: null,
     showSidePanel: true,
     orderedIdentifiers: [],
+    pastSidePanelState: {},
```
weili-govtech: `pastSidePanelState` and its corresponding reducer are no longer needed

src/components/side-panel/side-panel-header.tsx
```diff
@@ -29,32 +30,44 @@ export const SidePanelHeader = () => {
     //     if (onSaveChanges) onSaveChanges();
     //     return "Save button clicked";
     // };
-
     const handleCrossButtonClick = () => {
-        toggleMode(EBuilderMode.ADD_ELEMENT);
         removeFocusedElement();
-
+        toggleMode(currentMode);
+        togglePanel(showSidePanel);
```
weili-govtech: by passing the current states, these toggle functions are not updating anything. they should be removed

weili-govtech (changes requested):

weili-govtech (dismissed):

weili-govtech:

weili-govtech (approved):

PR #29, opened 2024-04-08
CCUBE-1218

ShIrInRuTh:
**Changes**
1. Merged the changes from CCUBE-1213 into this branch so as to use the mock context for the tests to be created for CCUBE-1218
2. Added new tests for the side-panel, side-panel-header & toolbar

**Additional information**

-   You may refer to this [CCUBE-1218](https://jira.ship.gov.sg/browse/CCUBE-1218)
-  ** Some of the tests in the side panel, side-header-panel test files may not work as they rely on the new chnages made in CCUBE-1202 and other branches that have yet to be merged to pass, so will need to run the tests again to check if they work oce the relevant changes have been merged to main**

PR #31, opened 2024-04-09
CCUBE-1221

ShIrInRuTh:
**Changes**
1. Created unit tests to test the conditional rendering of the main panel content 

**Additional information**

-   You may refer to this [CCUBE-1221](https://jira.ship.gov.sg/browse/CCUBE-1221)
- It has changes from CCUBE-1213 as i branched out from there and those changes havent yet been merged to the main branch

### DIFF
--- a/src/util/test-helper.tsx
+++ b/src/util/test-helper.tsx
@@ -8,7 +8,7 @@ import {
 } from "src/context-providers";
 import { ThemeProvider } from "styled-components";
 
-export const mockBuilderState: IBuilderState = {
+export const MOCK_BUILDER_STATE: IBuilderState = {
     mode: EBuilderMode.ADD_ELEMENT,
     elements: {},
     focusedElement: null,
@@ -29,7 +29,7 @@ export namespace TestHelper {
             <ThemeProvider theme={BaseTheme}>
                 <BuilderContext.Provider
                     value={{
-                        state: { ...mockBuilderState, ...builderContext },
+                        state: { ...MOCK_BUILDER_STATE, ...builderContext },
                         dispatch: noop,
                     }}
                 >
--- a/tests/components/element-card/element-card.spec.tsx
+++ b/tests/components/element-card/element-card.spec.tsx
@@ -5,6 +5,18 @@ import { EElementType, TElement } from "src/context-providers";
 import { ELEMENT_BUTTON_LABELS } from "src/data/elements-data";
 import { TestHelper, MOCK_BUILDER_STATE } from "src/util/test-helper";
 
+const mockDeleteElement = jest.fn();
+
+jest.mock("src/context-providers/builder/hook.ts", () => {
+    const actual = jest.requireActual("src/context-providers/builder/hook.ts");
+    return {
+        useBuilder: () => ({
+            ...actual.useBuilder(),
+            deleteElement: mockDeleteElement,
+        }),
+    };
+});
+
 describe("element-card", () => {
     afterEach(() => {
         jest.restoreAllMocks();
@@ -34,7 +46,6 @@ describe("element-card", () => {
                 { element: MOCK_ELEMENT },
                 {
                     builderContext: {
-                        ...MOCK_BUILDER_STATE,
                         focusedElement: { element: MOCK_ELEMENT },
                     },
                 }
@@ -47,7 +58,6 @@ describe("element-card", () => {
                 { element: MOCK_ELEMENT },
                 {
                     builderContext: {
-                        ...MOCK_BUILDER_STATE,
                         focusedElement: {
                             element: MOCK_ELEMENT,
                             isDirty: true,
@@ -65,7 +75,6 @@ describe("element-card", () => {
                 { element: MOCK_ELEMENT },
                 {
                     builderContext: {
-                        ...MOCK_BUILDER_STATE,
                         focusedElement: { element: MOCK_ELEMENT },
                     },
                 }
@@ -122,15 +131,3 @@ const MOCK_ELEMENT = {
 };
 
 const mockOnClick = jest.fn();
-
-const mockDeleteElement = jest.fn();
-
-jest.mock("src/context-providers/builder/hook.ts", () => {
-    const actual = jest.requireActual("src/context-providers/builder/hook.ts");
-    return {
-        useBuilder: () => ({
-            ...actual.useBuilder(),
-            deleteElement: mockDeleteElement,
-        }),
-    };
-});
--- a/tests/components/element-card/element-card.spec.tsx
+++ b/tests/components/element-card/element-card.spec.tsx
@@ -3,7 +3,7 @@ import "jest-canvas-mock";
 import { ElementCard } from "src/components";
 import { EElementType, TElement } from "src/context-providers";
 import { ELEMENT_BUTTON_LABELS } from "src/data/elements-data";
-import { TestHelper, MOCK_BUILDER_STATE } from "src/util/test-helper";
+import { TestHelper } from "src/util/test-helper";
 
 const mockDeleteElement = jest.fn();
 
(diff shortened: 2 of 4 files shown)


src/util/test-helper.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { noop } from "lodash";
+import React from "react";
+import {
+    BuilderContext,
+    EBuilderMode,
+    IBuilderState,
+} from "src/context-providers";
+import { ThemeProvider } from "styled-components";
+
+export const mockBuilderState: IBuilderState = {
```
keithtxw: usually we put the mocks below the main exportable. Can add a section header stating `MOCKS`

And also rename it to be `MOCK_BUILDER_STATE`

tests/components/element-card/element-card.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { TestHelper, mockBuilderState } from "src/util/test-helper";
+
+describe("element-card", () => {
+    afterEach(() => {
+        jest.restoreAllMocks();
+        jest.resetAllMocks();
+    });
+
+    describe("onClick", () => {
+        it("should be in a focused state and show the edit element panel when clicked on", () => {
```
keithtxw: As you have no control of external, it is better to keep the test cases more relevant to the context of the component. 

```suggestion
        it("should fire the onClick callback when clicked", () => {
```

tests/components/element-card/element-card.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+                    builderContext: {
+                        ...mockBuilderState,
+                        focusedElement: { element: MOCK_ELEMENT },
+                    },
+                }
+            );
+            expect(getDuplicateButton()).toBeInTheDocument();
+            expect(getDeleteButton()).toBeInTheDocument();
+        });
+        it("should disable the duplicate button when there is already a duplicated element card", () => {
```
keithtxw: The test description doesn't really match
```suggestion
        it("should disable the duplicate button when the current element is in focus", () => {
```

tests/components/element-card/element-card.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        },
+                    },
+                }
+            );
+            expect(getDuplicateButton()).toBeInTheDocument();
+            expect(getDuplicateButton()).toBeDisabled();
+        });
+    });
+});
+
```
keithtxw, line 88: Add these test cases too
- delete button click event (check that the `deleteElement` in `useBuilder` is called with the current element's internal id)

keithtxw: You'll need to mock `useBuilder` for the second test. Here is an example how
```tsx
const mockDeleteElement = jest.fn();

jest.mock("src/context-providers/builder/hook.tsx", () => {
	const actual = jest.requireActual("src/context-providers/builder/hook.tsx");
	return {
		useBuilder: () => ({
			...actual.useBuilder(), // we don't want to mock `focusedElement`
			deleteElement: mockDeleteElement
		}),
	};
});
```

tests/components/main-panel/main-panel.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { ELEMENT_BUTTON_LABELS } from "src/data/elements-data";
+import { TestHelper, mockBuilderState } from "src/util/test-helper";
+
+describe("main-panel", () => {
+    afterEach(() => {
+        jest.restoreAllMocks();
+        jest.resetAllMocks();
+    });
+
+    describe("conditional rendering of main panel content", () => {
```
keithtxw: don't think we need this extra describe

tests/components/main-panel/main-panel.spec.tsx
```diff
@@ -0,0 +1,67 @@
+import { render, screen } from "@testing-library/react";
+import "jest-canvas-mock";
+import { MainPanel } from "src/components";
+import { EElementType } from "src/context-providers";
+import { ELEMENT_BUTTON_LABELS } from "src/data/elements-data";
+import { TestHelper, mockBuilderState } from "src/util/test-helper";
+
+describe("main-panel", () => {
```
keithtxw: We usually use the name of the component. In this case it will be `MainPanel`

tests/components/main-panel/main-panel.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    afterEach(() => {
```
keithtxw, line 9: No mocking is done. Hence no need for all these.

tests/components/element-card/element-card.spec.tsx
```diff
@@ -0,0 +1,109 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import "jest-canvas-mock";
+import { ElementCard } from "src/components";
+import { EElementType, TElement } from "src/context-providers";
+import { ELEMENT_BUTTON_LABELS } from "src/data/elements-data";
+import { TestHelper, mockBuilderState } from "src/util/test-helper";
+
+describe("element-card", () => {
```
keithtxw: We usually use the component's name. So here it should be `ElementCard`

tests/components/main-panel/main-panel.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+        jest.restoreAllMocks();
+        jest.resetAllMocks();
+    });
+
+    describe("conditional rendering of main panel content", () => {
+        it("should display the empty panel content", async () => {
```
keithtxw:
```suggestion
        it("should display the empty panel content if there are no elements in the panel", async () => {
```

tests/components/main-panel/main-panel.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+            renderComponent();
+            await expect(getEmptyPanelContent()).resolves.toBeInTheDocument();
```
keithtxw: I think can just get directly instead of having a function for it. I dont foresee any other test cases at the moment

keithtxw: Do you need the await? I don't think there are any state changes?

keithtxw (dismissed):

tests/components/main-panel/main-panel.spec.tsx
ShIrInRuTh: when i remove the async i get this error 
image

tests/components/element-card/element-card.spec.tsx
ShIrInRuTh: This has been changed in the PR for CCUBE-1218

src/util/test-helper.tsx
weili-govtech: sorry I need to interject on this, actually there is no need to export / spread `MOCK_BUILDER_STATE` in other tests because it will always be spread inside `TestHelper.withProviders()`, may I request to:
- not export `MOCK_BUILDER_STATE`
- remove using `MOCK_BUILDER_STATE` in all other tests

note: this is the same comment as https://github.com/LifeSG/web-form-builder/pull/29

tests/components/element-card/element-card.spec.tsx
```diff
@@ -107,3 +122,15 @@ const MOCK_ELEMENT = {
...
+
+jest.mock("src/context-providers/builder/hook.ts", () => {
+    const actual = jest.requireActual("src/context-providers/builder/hook.ts");
+    return {
+        useBuilder: () => ({
+            ...actual.useBuilder(),
+            deleteElement: mockDeleteElement,
+        }),
+    };
+});
```
weili-govtech: logically speaking, the mocks are hoisted to the top of the module it is mocking so we place them as close to the imports as possible. hence pls move this block of codes above the main test suite, just below the imports

src/util/test-helper.tsx
weili-govtech: this is not updated?

weili-govtech:

weili-govtech (approved):

PR #34, opened 2024-04-17
CCUBE-1223

ShIrInRuTh:
**Changes**
 - Created the save changes button
 - Allowed the tracking and saving of the field values
 - Created context providers to handle updating of the element & focused element
 - Amended error messages according to figma

**Additional information**

-   You may refer to this [CCUBE-1223](https://jira.ship.gov.sg/browse/CCUBE-1223)


src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -41,59 +51,75 @@ export const BasicDetails = () => {
...
+                            field.onChange({ target: { value: value } })
+                        }
                     />
                 )}
             />
+
             {hasProperty("label") && (
                 <Controller
                     name="label"
                     control={control}
```
weili-govtech, line 55: for controllers, add `shouldUnregister={true}` to remove field from react-hook-form automatically when field is unmounted
while this doesnt do anything now, it will be needed when we conditionally render the fields according to the field type selected

src/components/side-panel/side-panel-header.tsx
```diff
@@ -24,11 +29,12 @@ export const SidePanelHeader = () => {
...
 
-    // TODO: When react hook form is being set up, run this function only when there are changes in the form values
-    // const handleSaveButtonClick  = () => {
-    //     if (onSaveChanges) onSaveChanges();
-    //     return "Save button clicked";
-    // };
+    const handleSaveButtonClick = () => {
+        if (saveChangesHandler) {
+            saveChangesHandler();
+        }
```
weili-govtech: while this works fine, it is awkward to pass a function around. instead of this, may I propose to wrap the all the contents in the `SidePanel` in the `FormProvider` and `form`? this way, your submit button can submit the form out-of-the-box without any special handling

src/components/element-editor/element-editor.tsx
```diff
@@ -1,28 +1,80 @@
...
+            updateElement(values);
+            updateFocusedElement(false, values);
+            methods.reset(
+                {
+                    ...methods.getValues(),
+                },
+                {
+                    keepDirty: false,
+                }
+            );
```
weili-govtech: why do we need to reset to the latest form values after submission?
keepDirty is false by default too

src/components/element-editor/element-editor.tsx
```diff
@@ -1,28 +1,80 @@
...
     // =============================================================================
 
+    const renderAlert = () => {
+        if (focusedElement.isDirty === false) {
+            return <></>;
+        }
+        return (
+            <Alert type="warning" showIcon>
+                To reflect changes on preview, save changes first.
+            </Alert>
```
weili-govtech:
```suggestion
            <SaveChangesAlert type="warning" showIcon>
                To reflect changes on preview, save changes first.
            </SaveChangesAlert>
```

src/components/element-editor/element-editor.styles.tsx
```diff
@@ -1,8 +1,14 @@
...
 export const AccordionWrapper = styled(Accordion)`
+    margin-top: 1rem;
     & > div:first-child {
         display: none;
     }
 `;
+
+export const SaveChangesAlert = styled(Alert)`
+    margin: 1rem;
+`;
```
weili-govtech: layout does not match what is on figma:
- alert box should be shorter than accordion wrapper
- alert box should have vertical space of 1 rem

src/components/side-panel/side-panel-header.tsx
```diff
@@ -54,13 +60,20 @@ export const SidePanelHeader = () => {
     const renderIconButton = () => {
```
weili-govtech:
```suggestion
    const renderButtons = () => {
```

src/schemas/base-helper.ts
```diff
@@ -10,14 +10,14 @@ export namespace BaseSchemaHelper {
...
-            label: yup.string().required("Label is required"),
+            type: yup.string().required("Element type required"),
+            label: yup.string().required("Label required"),
             placeholder: yup.string().optional(),
             required: yup.boolean().required().default(true),
             requiredErrorMsg: yup.string().optional(),
             id: yup
                 .string()
-                .required("ID is required")
+                .required("ID required")
```
weili-govtech: missing period at the end of each error message

weili-govtech (changes requested):

src/components/element-editor/element-editor.tsx
ShIrInRuTh: for this case im resetting the form mainly for the isDirty prop cause it remains as true even after submission which will affect the rendering of the Save Changes button label and the rendering of the alert

src/components/side-panel/side-panel-header.tsx
```diff
@@ -18,21 +22,41 @@ export const SidePanelHeader = () => {
...
-    //     if (onSaveChanges) onSaveChanges();
-    //     return "Save button clicked";
-    // };
     const handleCrossButtonClick = () => {
         removeFocusedElement();
     };
 
+    const onSubmit = useCallback(
+        (values) => {
+            console.log("values:", values);
```
weili-govtech: errant console.log

src/components/element-editor/element-editor.styles.tsx
weili-govtech: layout still does not match what is on figma:
- missing gap below alert
- accordion content should have 1rem horizontal padding
- gap below "Basic" title should be 2rem
- gap below "Basic" title should be 1rem when accordion is collapsed

image
image

image

src/components/element-editor/basic-details/basic-details.tsx
weili-govtech, line 1: fields are not showing the updated values if I switch between elements

src/components/side-panel/side-panel.tsx
```diff
@@ -30,16 +38,18 @@ export const SidePanel = () => {
...
+                <ContentWrapper>
+                    <ContentSection
+                        $isFocusedElement={focusedElement ? true : false}
+                    >
+                        {renderPanelContent()}
+                    </ContentSection>
+                    {focusedElement === null && <Toolbar />}
+                </ContentWrapper>
+            </Wrapper>
+        </FormProvider>
```
weili-govtech: - wrap the contents in `<form>` and have it handle the submit event
- this way, the submit button in `SidePanelHeader` just need to prevent html submission and does not need to handle the submit event

reason for handling here is because the bulk of the react-hook-form integration is here and the logic will be more visible here

```suggestion
        <FormProvider {...methods}>
            <form onSubmit={handleSubmit(onSubmit)()}>
                <Wrapper $minimised={focusedElement ? false : !showSidePanel}>
                    <SidePanelHeader />
                    <ContentWrapper>
                        <ContentSection
                            $isFocusedElement={focusedElement ? true : false}
                        >
                            {renderPanelContent()}
                        </ContentSection>
                        {focusedElement === null && <Toolbar />}
                    </ContentWrapper>
                </Wrapper>
            </form>
        </FormProvider>
```

src/components/side-panel/side-panel-header.tsx
```diff
@@ -18,21 +22,41 @@ export const SidePanelHeader = () => {
...
+            updateElement(values);
+            updateFocusedElement(false, values);
+            reset(
+                {
+                    ...values,
+                },
+                {
+                    keepDirty: false,
+                }
+            );
```
weili-govtech: can follow the implementation here: https://stackoverflow.com/a/76741818
but implement this in `SidePanel` instead
so the submit handler here should be preventing the native form submission via `event.preventDefault()`

```
const {
    formState: { isSubmitSuccessful },
    reset,
    handleSubmit,

useEffect(() => {
    reset(undefined, {
        keepValues: true,
        keepDirty: false,
    });
}, [isSubmitSuccessful]);
```

weili-govtech (changes requested):

weili-govtech (changes requested):
I noticed the `ElementEditor` is broken now and we have no use for it anymore, may I request to remove the story

src/components/element-editor/element-editor.styles.tsx
weili-govtech: there should be a 1rem spacing to the right of the arrow

image
image

src/components/side-panel/side-panel.tsx
```diff
@@ -9,8 +14,52 @@ export const SidePanel = () => {
...
+            updateElement(values);
+            updateFocusedElement(false, values);
+        },
+        [updateElement, updateFocusedElement]
+    );
+    // =========================================================================
+    // USE EFFECTS
+    // =========================================================================
+    useEffect(() => {
+        console.log("check this: ", methods.formState.isSubmitSuccessful);
```
weili-govtech: errant console.log

src/components/element-editor/basic-details/basic-details.tsx
weili-govtech: I notice the fields are not showing the correct values if I switch between fields with unsaved changes (aka is dirty)

weili-govtech:
need to resolve merge conflicts too

weili-govtech (changes requested):

weili-govtech (approved):

PR #39, opened 2024-04-23
CCUBE-1263

ShIrInRuTh:
**Changes**
- Creating the drag & drop functionality
- Amending the onClick test case for the element-card
- Adding a test case to test the rendering of the drag handle when hovering over the element card

**Additional information**

-   You may refer to this [CCUBE-1263](https://jira.ship.gov.sg/browse/CCUBE-1263)

### DIFF
--- a/src/components/common/base-card/base-card.styles.tsx
+++ b/src/components/common/base-card/base-card.styles.tsx
@@ -14,6 +14,7 @@ interface IBodyStyleProps {
 export const Body = styled.div<IBodyStyleProps>`
     padding: 1rem;
     border-radius: 0.5rem;
+    cursor: pointer;
     width: 100%;
     position: relative;
     display: flex;
--- a/src/components/common/base-card/base-card.styles.tsx
+++ b/src/components/common/base-card/base-card.styles.tsx
@@ -13,7 +13,6 @@ interface IBodyStyleProps {
 // =============================================================================
 export const Body = styled.div<IBodyStyleProps>`
     padding: 1rem;
-    cursor: pointer;
     border-radius: 0.5rem;
     width: 100%;
     position: relative;
--- a/src/components/element-card/element-card.styles.tsx
+++ b/src/components/element-card/element-card.styles.tsx
@@ -94,11 +94,15 @@ export const DroppableWrapper = styled.div<IDroppableWrapperProps>`
                   border-radius: 0.25rem;
                   background: ${Color.Accent.Light[5]};
                   height: 100%;
+                  z-index: -1;
               `
             : css`
                   border: 1px solid transparent;
               `}
-    height: 100%;
+    position: absolute;
+    flex-direction: column;
+    height: auto;
+    width: 100%;
     display: flex;
     flex-direction: column;
     align-items: center;
--- a/src/components/element-card/element-card.styles.tsx
+++ b/src/components/element-card/element-card.styles.tsx
@@ -9,15 +9,31 @@ import styled, { css } from "styled-components";
 interface IActionButtonStyleProps {
     $disabled?: boolean;
 }
+interface DroppableWrapperProps {
+    isOver: boolean;
+}
 
 // =============================================================================
 // STYLING
 // =============================================================================
+export const DragHandle = styled(DragHandleIcon)`
+    display: none;
+    transition: width 0.1s ease-out;
+
+    svg {
+        height: 1.25rem;
+        width: 1.25rem;
+    }
+`;
+
 export const Container = styled.div`
     display: flex;
     align-items: center;
     gap: 1rem;
     flex: 1;
+    &:hover ${DragHandle} {
+        display: block;
+    }
 `;
 
 export const DetailsContainer = styled.div`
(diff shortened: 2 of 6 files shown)


src/components/element-card/element-card.styles.tsx
```diff
@@ -68,3 +69,10 @@ export const ActionButton = styled.button<IActionButtonStyleProps>`
         }
     }}
 `;
+
+export const DragHandle = styled(DragHandleIcon)`
+    svg {
+        height: 1.25rem;
+        width: 1.25rem;
+    }
+`;
```
weili-govtech: instead of conditionally rendering the `DragHandle`, you can use `:hover` to apply hover effects, e.g.

```
export const DragHandle = styled(DragHandleIcon)`
  display: none;
  // ...
`;

export const Main = styled.div`
    &:hover ${DragHandle} {
        display: block;
    }
`;
```

ps: `Main` refers to the outermost div in ElementCard

weili-govtech: can also explore animating the width of `DragHandle` on hover via transition, e.g. `transition: width 0.1s ease-out;`

src/components/element-card/element-card.tsx
weili-govtech, line 1: notice some features are missing when compared against figma:
- while dragging, the cursor is not showing the grabbing cursor
- dragged element should be rendered in 0.7 opacity
- there is a `Drop your element here` placeholder to indicate where the dragged element will be dropped onto
- dragged element should not go beneath other elements; it should always have highest z-index

src/components/main-panel/main-panel.tsx
```diff
@@ -20,10 +37,41 @@ export const MainPanel = () => {
...
+        })
+    );
+
+    for (const orderedIdentifier of orderedIdentifiers) {
+        if ("internalId" in orderedIdentifier) {
+            items.push({ id: orderedIdentifier.internalId });
+        } else {
+            items.push(orderedIdentifier);
+        }
+    }
```
weili-govtech: either move this under `CONST, STATE, REFS` divider

or store items as a state and update it through useEffect under `EFFECTS` divider

weili-govtech (changes requested):

src/components/main-panel/main-panel.tsx
```diff
@@ -20,10 +38,42 @@ export const MainPanel = () => {
         elements,
         focusElement,
         focusedElement,
+        updateOrderedIdentifiers,
     } = useBuilder();
 
     const finalMode = focusedElement ? true : showSidePanel;
     const renderMode = finalMode ? "minimised" : "expanded";
+    const items: (UniqueIdentifier | { id: UniqueIdentifier })[] = [];
+    const [activeId, setActiveId] = useState<string>();
```
weili-govtech: remove as it is not used anymore

src/components/common/base-card/base-card.styles.tsx
```diff
@@ -13,7 +13,6 @@ interface IBodyStyleProps {
 // =============================================================================
 export const Body = styled.div<IBodyStyleProps>`
     padding: 1rem;
-    cursor: pointer;
```
weili-govtech, line 16: should keep this as by removing this style, all other buttons are showing the default arrow cursor.

I believe your component is not able to receive the styles when you convert it to a styled-component? in order to allow `BaseCard` to be a styled-component, it merely needs to receive the `className` prop

```
interface IProps {
  // ...
  className?: string;
}

export const BaseCard = ({ className, id, children, focused, onClick }: IProps) => {
    return (
        <Body
            tabIndex={0}
            $focused={focused}
            onClick={onClick}
            id={id}
            className={className}
            role="button"
        >
            {children}
        </Body>
    );
};

```

src/components/element-card/element-card.styles.tsx
```diff
@@ -68,3 +85,41 @@ export const ActionButton = styled.button<IActionButtonStyleProps>`
...
+        height: 2.08rem;
+        width: 2.08rem;
+    }
+`;
+
+export const DroppableText = styled(Text.Body)`
+    color: ${Color.Primary};
+    text-align: center;
+    font-size: 1rem;
+    font-weight: semi-bold;
```
weili-govtech: `semi-bold` is not a font-weight, use the `weight` prop to apply the font-weight instead

```
<DroppableText weight={...}>Drop your element here</DroppableText>
```

src/components/element-card/element-card.tsx
```diff
@@ -25,9 +32,34 @@ export const ElementCard = ({ element, onClick }: IProps) => {
...
     const isFocused = checkIsFocused();
     const disableDuplicate = shouldDisableDuplicate();
 
+    const style = {
+        transform: CSS.Translate.toString(transform),
+        transition,
+        opacity: isDragging ? "70%" : "100%",
+        background: isDragging ? "white" : "inherit",
+        gap: isDragging ? "1rem" : "inherit",
+        zIndex: isDragging ? 1 : "auto",
```
weili-govtech, line 57: zIndex is not working, need to apply `position` too

weili-govtech (changes requested):

src/components/element-card/element-card.tsx
```diff
@@ -25,9 +34,35 @@ export const ElementCard = ({ element, onClick }: IProps) => {
...
     const disableDuplicate = shouldDisableDuplicate();
 
+    const style: CSSProperties = {
+        transform: CSS.Translate.toString(transform),
+        transition,
+        opacity: isDragging ? "70%" : "100%",
+        background: isDragging ? "white" : "inherit",
+        gap: isDragging ? "1rem" : "inherit",
+        width: "100%",
+        position: isDragging ? "absolute" : "relative",
```
weili-govtech: absolute positioning will cause your dragged card to not take up any space too. this causes your droppable content to go under other cards:

image

so using relative positioning is good enough already

```suggestion
        position: "relative",
```

weili-govtech (approved):

PR #40, opened 2024-04-25
CCUBE-1265

ShIrInRuTh:
**Changes**
- Created multi entry component, 
- Created onAdd & delete functions
- Updated unit tests for validation-child 
- Ensured the values are saved in local state onChange, then deleted when component is deleted
- Created unit test to tes the validation component

**Additional information**

-   You may refer to this [CCUBE-1265](https://jira.ship.gov.sg/browse/CCUBE-1265)
-   There may be a quite a few changes linked to it cause its branched out from CCUBE-1264


src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,84 @@
...
+            );
+            if (prevValueWithSameType) {
+                return prevValue.map((value: IEntryProps) =>
+                    value.type === type ? { ...value, ...newValue } : value
+                );
+            } else {
+                return [...prevValue, newValue];
+            }
+        });
+    };
```
weili-govtech: we want to maintain a separation of concern between the parent and child component. this means that we should not pass a parent's state setter to the child.

the child should only need to pass the rule and errorMessage to the parent. it will be up to the parent to manage the update to its own internal state.

src/components/element-editor/validation/validation.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const Validation = () => {
+    // =========================================================================
+    // CONST, STATES, REFS
+    // =========================================================================
+    const { focusedElement } = useBuilder();
+    const element = focusedElement.element;
+    const [numberOfChildEntry, setNumberOfChildEntry] = useState(0);
+    const [childEntryValues, setChildEntryValues] = useState<IEntryProps[]>([]);
+    const [childKey, setChildKey] = useState([]);
```
weili-govtech: proposing to use `childEntryValues` to track the no. of entries so you can remove  `childKey` and `numberOfChildEntry`.

I see that `childKey` is used for the `key` prop, you can use something like `validation-entry-${index}` in place of it

src/components/element-editor/validation/validation.tsx
```diff
@@ -0,0 +1,103 @@
...
+            );
+            const updatedKeys = childKey.filter(
+                (_, index) => index !== childEntryValues.indexOf(childToRemove)
+            );
+
+            setChildEntryValues(updatedValues);
+            setChildKey(updatedKeys);
+            setNumberOfChildEntry((num) => num - 1);
+        }
+    };
```
weili-govtech: - rename as `handleDelete`
- instead of type, pass index as argument, you can remove the childEntryValue by the index

weili-govtech (changes requested):
note: I will probably have more feedbacks because I understand that this PR is behind a few other PRs in review. I will review in more detail once the rest of the PRs are resolved.

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+    // CONST, STATE, REF
+    // =========================================================================
+    const [type, setType] = useState<string>();
+
+    // =============================================================================
+    // EVENT HANLDERS
+    // =============================================================================
+    const handleChange = (changeType?: string, newValue?: string) => {
+        const updatedValue = { ...value };
+        if (!updatedValue.validationType || options.length === 1) {
```
weili-govtech: this wont work if there are multiple validation types, why not use just set the type directly when `Form.Select` has a change in selected option?

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const ValidationChild = ({
+    onDelete,
+    options,
+    onChange,
+    value,
+}: IProps) => {
+    // =========================================================================
+    // CONST, STATE, REF
+    // =========================================================================
+    const [type, setType] = useState<string>();
```
weili-govtech: why do we need this state?

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <div>
+                    <Form.Select
+                        placeholder="Select"
+                        selectedOption={type}
+                        options={options}
+                        disabled={options.length === 1}
+                        onSelectOption={(option) => {
+                            if (options.length > 1) setType(option as string);
+                        }}
+                        onBlur={handleChange}
```
weili-govtech: why the need to call handleChange on blur? why not call handleChange in `onSelectOption`?

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+    // USE EFFECTS
+    // =========================================================================
+
+    useEffect(() => {
+        if (options.length === 1) {
+            setType(options[0]);
+        } else {
+            setType(value.validationType);
+        }
+    }, [options]);
```
weili-govtech: - why not call `handleChange` to pass the validationType to parent?
- should default to 1st option if there is only 1 option AND `value.validationType` is not set
- else block is probably not needed if we are doing away with storing type in local state

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        }}
+                        onBlur={handleChange}
+                    />
+                </div>
+                <div>
+                    <Form.Textarea
+                        placeholder="Enter rule"
+                        defaultValue={value.validationRule}
+                        onChange={(event) => {
+                            event.preventDefault();
```
weili-govtech: what default behaviours do we need to stop?

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,96 @@
...
+                            handleChange("rule", event.target.value);
+                        }}
+                    />
+                </div>
+                <div>
+                    <Form.Input
+                        placeholder="Set error message"
+                        defaultValue={value.validationErrorMessage}
+                        onChange={(event) => {
+                            event.preventDefault();
```
weili-govtech: what default behaviours do we need to stop?

src/components/element-editor/validation/validation.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    function getMaxEntries(elementType: EElementType) {
+        switch (elementType) {
+            case EElementType.EMAIL:
+                return ELEMENT_VALIDATION_TYPES["Text field"][
+                    EElementType.EMAIL
+                ].maxEntries;
+            default:
+                return 6;
```
weili-govtech: is this an arbitrary number? if so, can add a comment to indicate that

src/components/element-editor/validation/validation.tsx
```diff
@@ -0,0 +1,107 @@
...
+                key={`validation-entry-${index}`}
+                onDelete={() => handleDelete(index)}
+                onChange={(newValue) => handleChildChange(index, newValue)}
+                options={getOptionsByType(element.type)}
+                value={child}
+            />
+        ));
+    }, [childEntryValues, handleDelete, handleChildChange, element.type]);
+
+    console.log("check child entry:", childEntryValues);
```
weili-govtech: errant console.log

src/data/elements-data.ts
```diff
@@ -31,3 +31,14 @@ export const ELEMENT_BUTTON_LABELS: { [key in EElementType]: string } = {
     [EElementType.CHECKBOX]: "Checkbox",
     [EElementType.RADIO]: "Radio Button",
 };
+// TODO: change the array to an object specifying the maxValidation entires and then the validation types
+//  E.g. [EElementType.EMAIL]:{ maxEntries: 1, validationTypes: ["Valid email address format"]},
+
```
weili-govtech: can remove

tests/components/element-editor/validation/validation-child.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+            onChange: mockOnChange,
+        });
+        expect(screen.getByText("Select")).toBeInTheDocument();
+        expect(screen.getByPlaceholderText("Enter rule")).toBeInTheDocument();
+        expect(
+            screen.getByPlaceholderText("Set error message")
+        ).toBeInTheDocument();
+    });
+
+    it("should run the delete function when clicking on the bin icon", () => {
```
weili-govtech:
```suggestion
    it("should fire the delete function on clicking the bin icon", () => {
```

tests/components/element-editor/validation/validation-child.spec.tsx
```diff
@@ -0,0 +1,91 @@
...
+            onChange: mockOnChange,
+        });
+        const getValidationTypeField = screen.getByRole("button", {
+            name: "Option 1",
+        });
+        expect(getValidationTypeField).toBeInTheDocument();
+        expect(getValidationTypeField).toBeDisabled();
+    });
+
+    it("should run onChange when there is a change in the input fields", () => {
```
weili-govtech:
```suggestion
    it("should fire onChange when there is a change in the input fields", () => {
```

tests/components/element-editor/validation/validation.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+        jest.resetAllMocks();
+    });
+
+    it("should contain the the component with the title, buttonLabel & children being passed into it", () => {
+        renderComponent({
+            builderContext: {
+                focusedElement: { element: MOCK_ELEMENT },
+            },
+        });
+        console.log(screen.debug());
```
weili-govtech: errant console.log

tests/components/element-editor/validation/validation.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+        expect(screen.getByText("Validation")).toBeInTheDocument();
+        expect(screen.getByText("Add validation")).toBeInTheDocument();
+    });
+
+    it("should run onAdd and render the validation-child component when the button is being clicked", () => {
```
weili-govtech:
```suggestion
    it("should fire onAdd and render the validation-child component when the button is being clicked", () => {
```

tests/components/element-editor/validation/validation.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+        renderComponent({
+            builderContext: {
+                focusedElement: { element: MOCK_ELEMENT },
+            },
+        });
+        console.log(screen.debug());
```
weili-govtech: errant console.log

weili-govtech (changes requested):

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -0,0 +1,92 @@
+import { Form } from "@lifesg/react-design-system";
```
weili-govtech: use subdirectory import to reduce bundled size

```suggestion
import { Form } from "@lifesg/react-design-system/form";
```

weili-govtech (changes requested):

src/components/element-editor/validation/validation.tsx
```diff
@@ -0,0 +1,105 @@
...
+        return childEntryValues?.map((child, index) => (
+            <ValidationChild
+                key={`validation-entry-${index}`}
+                onDelete={() => handleDelete(index)}
+                onChange={(newValue) => handleChildChange(index, newValue)}
+                options={getOptionsByType(element.type)}
+                value={child}
+            />
+        ));
+    }, [childEntryValues, handleDelete, handleChildChange, element.type]);
```
weili-govtech: actually there is no need for `useCallback` because that merely memo-ise the function; it will still run on every render

weili-govtech (changes requested):

weili-govtech (approved):

PR #43, opened 2024-05-07
[CCUBE-1223][SR] Amended styles for the accordion item for the border not to be displayed when the alert is not showing

ShIrInRuTh:
**Changes**

- Amended styles for the accordion item for the border not to be displayed when the alert is not showing

### DIFF
--- a/src/components/element-editor/basic-details/basic-details.styles.tsx
+++ b/src/components/element-editor/basic-details/basic-details.styles.tsx
@@ -2,8 +2,8 @@ import { Color } from "@lifesg/react-design-system/color";
 import { Accordion } from "@lifesg/react-design-system/accordion";
 import styled from "styled-components";
 
-interface IProps {
-    $displayAlert: boolean;
+interface FieldEditorAccordionItemProps {
+    $hideTopBorder: boolean;
 }
 
 export const MandatoryFieldBox = styled.div`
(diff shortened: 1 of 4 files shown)


src/components/element-editor/basic-details/basic-details.styles.tsx
```diff
@@ -1,6 +1,11 @@
+import { Color } from "@lifesg/react-design-system/color";
 import { Accordion } from "@lifesg/react-design-system/accordion";
 import styled from "styled-components";
 
+interface IProps {
```
weili-govtech: since there are multiple components here, the name of the typing will need to be more specific

```suggestion
interface FieldEditorAccordionItemProps {
```

src/components/element-editor/basic-details/basic-details.styles.tsx
```diff
@@ -1,6 +1,11 @@
...
+    $displayAlert: boolean;
```
weili-govtech: since the component is not rendering the alert, can rename it to be more specific to what it is doing
```suggestion
    $hideTopBorder: boolean;
```

weili-govtech (changes requested):
- when the alert is not rendered, can remove the spacing between the header and the accordion should be 1rem only
- arrow from validation accordion is not aligned

image

weili-govtech:
> * when the alert is not rendered, can remove the spacing between the header and the accordion should be 1rem only
> * arrow from validation accordion is not aligned
> 
> image

this is not fixed?

weili-govtech:
image

need to vertically center the arrow icon too

weili-govtech:
arrow is still misaligned when the panel is expanded
image

weili-govtech (approved):

PR #44, opened 2024-05-07
[CCUBE-1225][SR] Amending button styles and changing it to use the ButtonWithIcon component.

ShIrInRuTh:
**Changes**
- Amending button styles and changing it to use the ButtonWithIcon component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
         "@babel/preset-env": "^7.16.11",
         "@babel/preset-react": "^7.16.7",
         "@babel/preset-typescript": "^7.16.7",
-        "@lifesg/react-design-system": "^2.4.0-canary.1",
+        "@lifesg/react-design-system": "^2.5.0",
         "@lifesg/react-icons": "^1.4.0",
         "@rollup/plugin-commonjs": "^25.0.7",
         "@rollup/plugin-image": "^3.0.3",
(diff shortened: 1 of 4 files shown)


package.json
weili-govtech, line 1: since design system has been updated, the changes to package-lock should be included too.
we want to include package-lock because it lists the actual packages to be installed.

src/components/common/multi-entry/multi-entry.tsx
```diff
@@ -26,15 +24,15 @@ export const MultiEntry = ({
         <MultiEntryAccordionItem title={title}>
             <>
                 {children}
-                <AddMultiEntryButton
+                <ButtonWithIcon.Default
```
weili-govtech: - button needs 100% width
- button shadow is getting cropped off by accordion, it used to have 2rem margin and that prevented the cropping

weili-govtech (changes requested):

src/components/common/multi-entry/multi-entry.tsx
weili-govtech: thanks for the change but the update causes it not to match the Figma mocks, I think you need to have
- 1rem top and bottom margins for the button
- reduce the accordion item bottom margin to 1rem

weili-govtech (approved):

PR #45, opened 2024-05-08
CCUBE-930

ShIrInRuTh:
**Changes**

- Adding react hook form to the validation component
- Allowing the submission of the validation component and its field inputs via RHF
- Creating validation schema for the validation entry component fields 
- Ensuring validation for the fields occur when needed via RHF
- Updated the unit tests for both validation child & validation components so the tests will not break by adding Form Provider to it
- Added new tests in validation child to test the rendering of prefilled values and rendering of error messages when needed

**Additional information**

-   You may refer to this [CCUBE-930](https://jira.ship.gov.sg/browse/CCUBE-930)

### DIFF
--- a/src/components/element-editor/basic-details/basic-details.tsx
+++ b/src/components/element-editor/basic-details/basic-details.tsx
@@ -58,7 +58,7 @@ export const BasicDetails = () => {
                         <IconDropdown
                             type={element?.type}
                             id={element?.id}
-                            errorMessage={errors.type?.message}
+                            errorMessage={errors.type?.message as string}
                             onChange={(value) =>
                                 field.onChange({ target: { value: value } })
                             }
@@ -96,7 +96,7 @@ export const BasicDetails = () => {
                             <TogglePair
                                 label="Mandatory field"
                                 defaultValue={element.required}
-                                id={element.internalId}
+                                id={element.internalId as string}
                                 onChange={(value) => field.onChange(value)}
                             />
                         )}
--- a/src/components/element-editor/basic-details/basic-details.tsx
+++ b/src/components/element-editor/basic-details/basic-details.tsx
@@ -58,7 +58,7 @@ export const BasicDetails = () => {
                         <IconDropdown
                             type={element?.type}
                             id={element?.id}
-                            errorMessage={errors.type?.message as string}
+                            errorMessage={errors.type?.message}
                             onChange={(value) =>
                                 field.onChange({ target: { value: value } })
                             }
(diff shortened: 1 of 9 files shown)


src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -58,7 +58,7 @@ export const BasicDetails = () => {
                         <IconDropdown
                             type={element?.type}
                             id={element?.id}
-                            errorMessage={errors.type?.message}
+                            errorMessage={errors.type?.message as string}
```
weili-govtech: why the need to cast as string? isn't it a string already?

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -96,7 +96,7 @@ export const BasicDetails = () => {
                             <TogglePair
                                 label="Mandatory field"
                                 defaultValue={element.required}
-                                id={element.internalId}
+                                id={element.internalId as string}
```
weili-govtech: same question, why need to cast as string?

src/schemas/text-based-fields.ts
```diff
@@ -1,3 +1,34 @@
...
+            validationRule: yup.string().when("validationType", {
+                is: ELEMENT_VALIDATION_TYPES["Text field"][EElementType.EMAIL]
+                    .validationTypes[0],
+                then: (rule) =>
+                    rule
+                        .required("Email address required.")
+                        .matches(
+                            DOMAIN_REGEX,
+                            "Invalid email address format.  Check if email address is correct with no whitespace between characters. Separate each address with a comma if there is more than 1 email."
+                        ),
```
weili-govtech: the error messages are not accurate, they should be referring to the domains and not email addresses, may I trouble you to discuss with UX on this?

src/components/element-editor/validation/validation-child.tsx
weili-govtech, line 1: noticed the error messages are not updated upon submission, steps to reproduce
1. add element and go to edit details panel
2. add a validation rule without filling any fields
3. click "Save changes" to cause the error messages to show
4. edit the validation rule

the error messages are not getting updated thereafter

weili-govtech (changes requested):

src/schemas/text-based-fields.ts
ShIrInRuTh: have discussed will change it :)

weili-govtech (approved):

PR #51, opened 2024-05-16
[CCUBE-1293][SR] Amended the drop down styles, amended bug to save conditional rendering as an empty array when saving changes, ammending unit tests according to the bug fixes.

ShIrInRuTh:
**Changes**
- Amended the dropdown styles
- Amended bug to save conditional rendering as an empty array when saving changes
- Ammending unit tests according to the bug fixes.

**Additional information**

- It fixes the bug in the dev env, when saving changes in the edit element panle, the conditional rendering prop is left undefined causing it to not render new entries after saving and throwing an undefined error.
- It also fixes the condition dropdown so that the width is longer to show the dropdown list properly and according to figma.

### DIFF
--- a/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
+++ b/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
@@ -5,6 +5,8 @@ import {
     ConditionalRenderingChild,
     IOnChangeProps,
 } from "./conditional-rendering-child";
+import { useFormContext } from "react-hook-form";
+import { IBaseTextBasedFieldValues } from "src/schemas";
 
 interface IOptions {
     label: string;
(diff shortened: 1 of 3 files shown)


src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -21,6 +23,7 @@ export const ConditionalRendering = () => {
     const element = focusedElement.element;
     const [childEntryValues, setChildEntryValues] =
         useState<IConditionalRendering[]>();
```
weili-govtech, line 25: default to array instead of setting empty array on line 52
```suggestion
        useState<IConditionalRendering[]>([]);
```

weili-govtech (changes requested):

PR #52, opened 2024-05-16
CCUBE-1294

ShIrInRuTh:
**Changes**
- Amending duplicate button to be enabled when for is not dirty
- Creates hook and reducer for duplicate functionality
- Created new typing for the reducer
- Amended the regex to allow kebab case also
- Created a duplicate generator to generate duplicated elements
- Created a new unit test for duplicate functionality.

**Additional information**

-   You may refer to this [CCUBE-1294](https://jira.ship.gov.sg/browse/CCUBE-1294)

### DIFF
--- a/src/components/element-card/element-card.tsx
+++ b/src/components/element-card/element-card.tsx
@@ -5,8 +5,9 @@ import { Text } from "@lifesg/react-design-system/text";
 import { PlusCircleIcon } from "@lifesg/react-icons";
 import { BinIcon } from "@lifesg/react-icons/bin";
 import { CopyIcon } from "@lifesg/react-icons/copy";
+import { CSSProperties } from "react";
 import { TElement, useBuilder } from "src/context-providers";
-import { BaseCard, CardIcon } from "../common";
+import { CardIcon } from "../common";
 import {
     ActionButton,
     ActionsContainer,
@@ -18,7 +19,6 @@ import {
     ElementBaseCard,
     IdLabel,
 } from "./element-card.styles";
-import { CSSProperties } from "react";
 
 interface IProps {
     element: TElement;
@@ -32,7 +32,7 @@ export const ElementCard = ({ element, onClick }: IProps) => {
     // CONST, STATE, REFS
     // =========================================================================
     const { label, id } = element;
-    const { focusedElement, deleteElement } = useBuilder();
+    const { focusedElement, deleteElement, duplicateElement } = useBuilder();
 
     const { isDragging } = useDraggable({
         id: element.internalId,
@@ -75,6 +75,7 @@ export const ElementCard = ({ element, onClick }: IProps) => {
             event.preventDefault();
         }
         // TODO: Add handling
+        duplicateElement(focusedElement.element);
     };
 
     const handleDeleteClick = (event: React.MouseEvent<HTMLButtonElement>) => {
@@ -94,10 +95,7 @@ export const ElementCard = ({ element, onClick }: IProps) => {
     }
 
     function shouldDisableDuplicate() {
-        return (
-            checkIsFocused() &&
-            (focusedElement.isDirty || !focusedElement.isValid)
-        );
+        return checkIsFocused() && focusedElement.isDirty;
     }
 
     // =========================================================================
--- a/src/components/element-card/element-card.tsx
+++ b/src/components/element-card/element-card.tsx
@@ -74,7 +74,6 @@ export const ElementCard = ({ element, onClick }: IProps) => {
         if (disableDuplicate) {
             event.preventDefault();
         }
-        // TODO: Add handling
         duplicateElement(focusedElement.element);
     };
 
--- a/src/context-providers/builder/hook.ts
+++ b/src/context-providers/builder/hook.ts
@@ -63,19 +63,19 @@ export const useBuilder = () => {
             const existingIdentifiers = state.orderedIdentifiers.map(
                 (elementId) => elementId.internalId
             );
-            const newElement = ElementObjectGenerator.duplicate(
+            const duplicatedElement = ElementObjectGenerator.duplicate(
                 element,
                 existingIdentifiers
             );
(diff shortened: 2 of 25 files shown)


src/components/element-card/element-card.tsx
```diff
@@ -75,6 +75,7 @@ export const ElementCard = ({ element, onClick }: IProps) => {
             event.preventDefault();
         }
         // TODO: Add handling
```
weili-govtech: remove comment

src/util/element-object-generator.ts
```diff
@@ -40,4 +41,35 @@ export namespace ElementObjectGenerator {
...
+                    return newId;
+                }
+                case id.includes("-copy"):
+                    return id + "-2";
+                default:
+                    return id + "-copy";
+            }
+        };
+
+        const baseAttributes: IBaseAttributes = {
```
weili-govtech: with `TElement`, you won't need to cast the output as `TElement` in the hook
```suggestion
        const duplicatedElement: TElement = {
```

src/util/element-object-generator.ts
```diff
@@ -40,4 +41,35 @@ export namespace ElementObjectGenerator {
...
+                    const incrementedNumber = number + 1;
+                    const newId =
+                        parts.slice(0, -1).join("-") + "-" + incrementedNumber;
+                    return newId;
+                }
+                case id.includes("-copy"):
+                    return id + "-2";
+                default:
+                    return id + "-copy";
+            }
```
weili-govtech: it should also be checking if `<id>-copy` and `<id>-copy-<number>` exists in `existingIds`

else if I duplicate `emailAddress` twice, I will have 2 instances of `emailAddress-copy`

src/util/element-object-generator.ts
```diff
@@ -40,4 +41,35 @@ export namespace ElementObjectGenerator {
                 return null;
         }
     };
+
+    export const duplicate = (element: TElement, existingIds: string[]) => {
```
weili-govtech: this function can have unit tests to cover various scenarios

src/context-providers/builder/hook.ts
```diff
@@ -58,6 +58,39 @@ export const useBuilder = () => {
...
+                { internalId: newElement.internalId },
+            ];
+
+            dispatch({
+                type: "duplicate-element",
+                payload: {
+                    element: newElement as TElement,
+                    orderedIdentifiers: newOrderedIdentifiers,
+                },
+            });
```
weili-govtech: `duplicate-element` action appears redundant because it is not doing anything different from `add-element` so you can use `add-element` instead

weili-govtech (changes requested):

src/util/element-object-generator.ts
weili-govtech: issue still remains, to replicate:
1. add email address element
2. assign an id, e.g. `email` and save it
3. duplicate element: `email-copy` is added correctly
4. duplicate the original `email` element again, `email-copy` is added instead of `email-copy-2`

when an element is duplicated, there is a need to:
- extract the underlying id, e.g. extract `email` out of `email-copy`
- check against all the existing elements to determine how many `email` and `email-copy-x` are there
- generate the correct unique id with the correct number suffix

there are edge cases with the number suffix to consider, to simplify things, we will add to the highest suffix:
- `email` and `email-copy-2` exists, generated id will be `email-copy-3`
- `email-copy` and `email-copy-3` exists, generated id will be `email-copy-4`

tests/util/element-object-generator.spec.tsx
weili-govtech, line 1: given the various scenarios to generate the id, you can open a new test suite under `duplicate` test suite to cover that

src/util/element-object-generator.ts
ShIrInRuTh: these changes are in the latest push

tests/util/element-object-generator.spec.tsx
ShIrInRuTh: these changes are in the latest push

src/util/element-object-generator.ts
```diff
@@ -41,4 +42,48 @@ export namespace ElementObjectGenerator {
...
+                const match = existingId?.match(regex);
+                if (match) {
+                    const suffix = match[1] ? parseInt(match[1], 10) : 1;
+                    if (suffix > maxSuffix) {
+                        maxSuffix = suffix;
+                    }
+                }
+            });
+
+            return `${baseId}-copy${maxSuffix + 1 > 1 ? `-${maxSuffix + 1}` : ""}`;
```
weili-govtech: move these functions out of duplicate so they are not created every time duplicate is called

src/util/element-object-generator.ts
```diff
@@ -41,4 +42,48 @@ export namespace ElementObjectGenerator {
...
+
+            existingIds.forEach((existingId) => {
+                const match = existingId?.match(regex);
+                if (match) {
+                    const suffix = match[1] ? parseInt(match[1], 10) : 1;
+                    if (suffix > maxSuffix) {
+                        maxSuffix = suffix;
+                    }
+                }
+            });
```
weili-govtech: map all suffixes and then find the max once with `Math.max()`
```suggestion
            const suffixes = existingIds.map((existingId) => {
                const match = existingId?.match(regex);
                if (match) {
                    return match[1] ? parseInt(match[1], 10) : 1;
                }
                return 0;
            });
            const maxSuffix = Math.max(0, ...suffixes);
```

weili-govtech (changes requested):

src/util/element-object-generator.ts
ShIrInRuTh: changes are made on the latest push

src/util/element-object-generator.ts
ShIrInRuTh: changes are made on the latest push

weili-govtech (approved):

PR #54, opened 2024-05-17
CCUBE-1296

ShIrInRuTh:
**Changes**
-  Add popover to the multi entry component
- Added a function to check the popover content to be displayed
- Added function in validation to check if the button and render popover when fields are empty, input is invalid & max number of entries have been reached.
- Added new tests to test is the pop over is visible
- Added test to check the popover content being shown 
- Amended earlier test to test whether the button is disabled.
- Added popover functionality for conditional rendering 
- Amended the functions to be more readable and to render in real-time.

**Additional information**
-   You may refer to this [CCUBE-1296](https://jira.ship.gov.sg/browse/CCUBE-1296)

### DIFF
--- a/src/components/common/multi-entry/multi-entry.tsx
+++ b/src/components/common/multi-entry/multi-entry.tsx
@@ -3,6 +3,7 @@ import {
     AddMultiEntryButton,
     MultiEntryAccordionItem,
 } from "./multi-entry.styles";
+import { PopoverTrigger } from "@lifesg/react-design-system/popover-v2";
 
 interface IProps {
     title?: string;
@@ -10,6 +11,7 @@ interface IProps {
     onAdd: () => void;
     children: React.ReactNode;
     disabledButton?: boolean;
+    popoverReason?: string;
 }
 
 export const MultiEntry = ({
@@ -18,14 +20,43 @@ export const MultiEntry = ({
     onAdd,
     children,
     disabledButton,
+    popoverReason,
 }: IProps) => {
     // =============================================================================
     // RENDER FUNCTIONS
     // =============================================================================
-    return (
-        <MultiEntryAccordionItem title={title}>
-            <>
-                {children}
+
+    const renderPopoverContent = () => {
+        switch (popoverReason) {
+            case "maxEntry":
+                return `Limit reached. To add new ${buttonLabel}, remove existing ones first`;
+            case "emptyOrInvalid":
+                return `To add new ${buttonLabel}, fill up existing ${buttonLabel} first.`;
+        }
+    };
+
+    const renderButton = () => {
+        if (disabledButton && popoverReason) {
+            return (
+                <PopoverTrigger
+                    popoverContent={renderPopoverContent()}
+                    trigger="hover"
+                    data-testid="add-button-popover"
+                >
+                    <AddMultiEntryButton
+                        icon={<PlusIcon />}
+                        styleType="light"
+                        onClick={onAdd}
+                        role="button"
+                        disabled={disabledButton}
+                        type="button"
+                    >
+                        {"Add " + buttonLabel}
+                    </AddMultiEntryButton>
+                </PopoverTrigger>
+            );
+        } else {
+            return (
                 <AddMultiEntryButton
                     icon={<PlusIcon />}
                     styleType="light"
@@ -36,6 +67,15 @@ export const MultiEntry = ({
                 >
                     {"Add " + buttonLabel}
                 </AddMultiEntryButton>
+            );
+        }
+    };
+
+    return (
+        <MultiEntryAccordionItem title={title}>
+            <>
+                {children}
+                {renderButton()}
             </>
         </MultiEntryAccordionItem>
     );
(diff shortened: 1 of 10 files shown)


src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -46,6 +51,47 @@ export const ConditionalRendering = () => {
...
+            return hasErrors && noTouchedFields && hasChildEntries;
+        }
+    }
+
+    function getPopoverReason() {
+        if (getTouchedAndErrorsFields()) {
+            return EPopoverReason.EMPTY_OR_INVALID;
+        } else if (getElementOptions().length === 0) {
+            return EPopoverReason.NO_CONDITION;
+        }
```
weili-govtech: it appears that the copy are breaking into next line, can you check whether it is possible to reduce the padding of the popover element?

src/components/common/multi-entry/multi-entry.tsx
```diff
@@ -20,16 +23,46 @@ export const MultiEntry = ({
...
+    const renderPopoverContent = () => {
+        switch (popoverReason) {
+            case EPopoverReason.MAX_ENTRY:
+                return `Limit reached. To add new ${buttonLabel}, remove existing ones first`;
+            case EPopoverReason.EMPTY_OR_INVALID:
+                return `To add new ${buttonLabel}, fill up existing ${buttonLabel} first.`;
+            case EPopoverReason.NO_CONDITION:
+                return `No conditional rendering available.`;
+        }
+    };
```
weili-govtech: actual popover reason copy should be defined in the corresponding validation / conditional rendering / prefill components

so MultiEntry should accept any copy and render the popover accordingly

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -46,6 +51,47 @@ export const ConditionalRendering = () => {
...
+
+            return hasErrors || incompleteFields;
+        } else {
+            const hasErrors = errors?.conditionalRendering?.length !== 0;
+            const noTouchedFields =
+                touchedFields?.conditionalRendering?.length === undefined;
+            const hasChildEntries = childEntryValues?.length > 0;
+
+            return hasErrors && noTouchedFields && hasChildEntries;
+        }
```
weili-govtech: instead of coming up with the validation manually, you should rely on the actual Yup validation schema since that is the source of truth

```
 SchemaHelper.buildSchema(focusedElement.element.type)
            .pick(["conditionalRendering"])
            .validate(childEntryValues);
```

for touched fields, regardless whether they are touched or not, that is redundant right?

weili-govtech (changes requested):

src/schemas/helper.ts
```diff
@@ -6,17 +6,24 @@ import { EMAIL_SCHEMA } from "./text-based-fields";
 
 export namespace SchemaHelper {
     export const buildSchema = (type: EElementType) => {
-        const { elements, focusedElement } = useBuilder();
+        try {
+            const { elements, focusedElement } = useBuilder();
```
weili-govtech, line 10: just to explain why we were facing the `Rendered more hooks than during the previous render.` error:

React has a rule that hooks need to be ran on every render. `useBuilder()` which is a hook so we were facing when it was called in `getTouchedAndErrorsFields()`. moving it out of that function fixed the issue.

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
weili-govtech, line 1: image

regarding this issue, it is due how the values are updated in `handleChildChange()`, by updating RHF's values within a state setter, it appears to violate React's state update and rendering sequence.

instead of updating RHF's values within a setState, it should be done outside:

```
const updatedValues = [...childEntryValues];
updatedValues[index] = newValue;
setChildEntryValues(updatedValues);
setValue("conditionalRendering", updatedValues);
```

src/components/element-editor/conditional-rendering/conditional-rendering-child.tsx
weili-govtech, line 1: image

`Form.Select` does not offer refs so it is causing this error. you can either apply the individual properties of `field` manually or exclude `ref` from `field` before spreading it.

ShIrInRuTh: completed this change in the last push

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -46,6 +54,46 @@ export const ConditionalRendering = () => {
...
+            } catch (error) {
+                if (
+                    error.errors.some((errorMessage: string | string[]) =>
+                        errorMessage.includes("required")
+                    )
+                ) {
+                    return true;
+                } else {
+                    return false;
+                }
```
weili-govtech:
```suggestion
                return error.errors.some((errorMessage: string | string[]) =>
                    errorMessage.includes("required")
                );
```

weili-govtech (changes requested):

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
ShIrInRuTh: changes are made on the latest push

weili-govtech (approved):

PR #58, opened 2024-05-27
CCUBE-1347

ShIrInRuTh:
**Changes**
- Added states to handle changes between inputs for basic details, validation, conditional rendering & prefill
- Added fieldWithoutRef value into field
- Removed trigger as fieldWithoutRef handles validation on blur
- Add fieldWithoutRef.onChange and setState when onChange of inputs
- Changed typing for prefillMode to allow empty string as an empty state.

**Additional information**

-   You may refer to this [CCUBE-1347](https://jira.ship.gov.sg/browse/CCUBE-1347)


weili-govtech:
still getting errors with uncontrolled elements:

image

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -24,6 +24,14 @@ export const BasicDetails = () => {
...
     } = useFormContext<IBaseTextBasedFieldValues>();
     const element = focusedElement.element;
+    const [elementType, setElementType] = useState<EElementType>(element.type);
+    const [elementName, setElementName] = useState<string>(element.label);
+    const [required, setRequired] = useState<boolean>(element.required || true);
+    const [requiredErrorMsg, setRequiredErrorMsg] = useState<string>(
+        element.requiredErrorMsg || ""
+    );
+    const [id, setId] = useState<string>(element.id || "");
+    const [placeholder, setPlaceholder] = useState(element.placeholder || "");
```
weili-govtech: having a separate state to track each field may work if there are less fields but it gets messy with 5 fields now, can we create a single local state to store all the values as an object?

do the same for the rest of the components too

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -75,14 +84,19 @@ export const BasicDetails = () => {
                     <Controller
                         name="label"
                         control={control}
-                        defaultValue={element?.label}
+                        defaultValue={elementName}
```
weili-govtech: now that the fields are controlled, there is no need to set `defaultValue`

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -57,15 +65,16 @@ export const BasicDetails = () => {
...
+                    defaultValue={elementType}
                     render={({ field }) => (
                         <IconDropdown
-                            type={element?.type}
+                            type={elementType}
                             id={element?.id}
+                            onChange={(value: EElementType) => {
+                                setElementType(value);
+                                field.onChange({ target: { value: value } });
+                            }}
```
weili-govtech: notice each field makes a similar call to update the local state and fire an on change event, can we have a common function to do this?

weili-govtech (changes requested):

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in latest push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in latest push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in latest push

src/components/common/toggle-pair/toggle-pair.tsx
```diff
@@ -5,23 +5,27 @@ import { Row } from "./toggle-pair.styles";
 
 interface Props {
```
weili-govtech:
```suggestion
interface IProps {
```

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -1,17 +1,26 @@
...
-import { useBuilder } from "src/context-providers";
+import { EElementType, useBuilder } from "src/context-providers";
 import { IBaseTextBasedFieldValues } from "src/schemas";
 import {
     FieldEditorAccordionItem,
     MandatoryFieldBox,
     Wrapper,
 } from "./basic-details.styles";
 
+interface BasicDetailsProps {
```
weili-govtech:
```suggestion
interface IBasicDetailsProps {
```

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -33,6 +50,16 @@ export const BasicDetails = () => {
         return key in element;
     };
 
+    const onChangeField = (
+        value: string | EElementType | boolean,
+        fieldType: string,
+        field?: { onChange: (arg0: string | boolean) => void }
+    ) => {
+        const updatedValue = { ...basicDetails, [fieldType]: value };
+        setBasicDetails(updatedValue as BasicDetailsProps);
```
weili-govtech:
```suggestion
        setBasicDetails(updatedValue);
```

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -33,6 +50,16 @@ export const BasicDetails = () => {
         return key in element;
     };
 
+    const onChangeField = (
```
weili-govtech:
```suggestion
    const handleChangeField = (
```

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -33,6 +50,16 @@ export const BasicDetails = () => {
...
+        value: string | EElementType | boolean,
+        fieldType: string,
+        field?: { onChange: (arg0: string | boolean) => void }
```
weili-govtech: why is it optional?
instead of passing the entire field, the argument can just be the `onChange` handler

src/components/common/toggle-pair/toggle-pair.tsx
weili-govtech, line 1: should add commented dividers to separate different sections of the component

weili-govtech (changes requested):

src/components/common/toggle-pair/toggle-pair.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in the last push

src/components/common/toggle-pair/toggle-pair.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/validation/validation-child.tsx
```diff
@@ -2,67 +2,35 @@ import { Form } from "@lifesg/react-design-system/form";
...
-    onChange,
-    value,
-    index,
-}: IProps) => {
+export const ValidationChild = ({ onDelete, options, index }: IProps) => {
     const {
         formState: { errors },
         control,
-        trigger,
+        getValues,
```
weili-govtech: remove as it is not used

src/components/element-editor/validation/validation.tsx
```diff
@@ -55,10 +55,9 @@ export const Validation = () => {
...
             try {
                 const validationSchema = schema.pick(["validation"]);
-                const validationValues = getValues("validation");
 
                 validationSchema.validateSync({
-                    validation: validationValues,
+                    validation: childEntryValues,
                     abortEarly: false,
                 });
                 return false;
```
weili-govtech: let's try to use the values from RHF directly, store the values in a local variable and then use it through this component

`childEntryValues`'s role now is only to trigger a re-render, when a re-render happens, the local variable in this component will be refreshed with the latest values

weili-govtech:
error is still happening
image

weili-govtech (changes requested):

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -80,89 +75,87 @@ export const ConditionalRendering = () => {
...
         setValue("conditionalRendering", updatedChildEntries);
-        setChildEntryValues(updatedChildEntries);
-    }, [element]);
-
-    useEffect(() => {
-        setValue("conditionalRendering", childEntryValues, {
-            shouldDirty: true,
-        });
-    }, [childEntryValues]);
+    }, [element?.internalId]);
```
weili-govtech: why still need to set value manually?

src/components/element-editor/validation/validation.tsx
weili-govtech, line 1: notice validation not prefilling when I reopen the edit details panel

src/components/element-editor/validation/validation.tsx
weili-govtech: this is not done yet?

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -26,21 +23,22 @@ export const ConditionalRendering = () => {
...
-        childEntryValues?.length >
-            focusedElement?.element?.conditionalRendering?.length;
 
     const schema = SchemaHelper.buildSchema(EElementType.EMAIL);
     const invalidAndEmptyFields = getTouchedAndErrorsFields();
+    const conditionalRenderingValues = getValues("conditionalRendering") || [];
+    const shouldUpdateFocusedElement =
+        isDirty ||
+        conditionalRenderingValues?.length >
+            focusedElement?.element?.conditionalRendering?.length;
```
weili-govtech: can we update the focused element's dirty state based on `isDirty` flag alone?

I notice the form is not dirty on adding / removing entry, you can use `{shouldDirty: true}` in `setValue()` since the form has indeed been updated

if we can, this useEffect is not necessary since `BasicDetails` is already updating based on the dirty state

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -61,12 +59,12 @@ export const ConditionalRendering = () => {
     };
 
     function getTouchedAndErrorsFields() {
```
weili-govtech: apologies I think there is no need to check the length too and we can rename the function

```
function checkIsValid() {
        try {
            const validationSchema = schema.pick(["validation"]);

            validationSchema.validateSync({
                validation: validationValues,
                abortEarly: false,
            });
            return false;
        } catch (error) {
            return Yup.ValidationError.isError(error);
        }
    }
```

weili-govtech (changes requested):

src/components/element-editor/prefill/prefill.tsx
weili-govtech, line 1: remove unused `useBuilder` import

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -40,6 +40,8 @@ export const BasicDetails = () => {
     useEffect(() => {
         if (isDirty) {
             updateFocusedElement(isDirty);
+        } else {
+            updateFocusedElement(false);
         }
```
weili-govtech:
```suggestion
        updateFocusedElement(!!isDirty);
```

src/components/element-editor/validation/validation.tsx
```diff
@@ -112,13 +100,16 @@ export const Validation = () => {
...
     };
 
     const handleDelete = (index: number) => {
         const currentValues = [...validationValues];
         const updatedValues = currentValues.filter((_, i) => i !== index);
-        setValue("validation", updatedValues);
+        setValue("validation", updatedValues, { shouldDirty: true });
+        setTimeout(() => {
+            setValue("validation", updatedValues, { shouldDirty: true });
+        });
```
weili-govtech, line 117: add this comment above the first setValue so other devs wont be confused

```
/** * shouldDirty will only dirty the field; the dirty state is not propagated to the form level * workaround is to wait for RHF to register the change and set the value again * reference: https://github.com/orgs/react-hook-form/discussions/9913#discussioncomment-4936301 */
```

weili-govtech (changes requested):

src/components/element-editor/validation/validation.tsx
ShIrInRuTh: updated in latest changes

src/components/element-editor/validation/validation.tsx
ShIrInRuTh: updated in the latest changes

src/components/element-editor/prefill/prefill.tsx
ShIrInRuTh: updated in latest changes

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: updated in latest changes

weili-govtech (approved):

PR #59, opened 2024-05-27
CCUBE-1348

ShIrInRuTh:
**Changes**

- Added a new function to generate id in generate function for element-object-generator
- Added new tests for generate function
- Ammended hook function when adding new elements to include existingElements so that it can be checked when generating a new id.

**Additional information**

-   You may refer to this [CCUBE-1348](https://jira.ship.gov.sg/browse/CCUBE-1348)
- This branch is branched out from CCUBE-1294
- There is only one commit for this the rest is changes merged from main & commits from CCUBE-1294 since its branched from there


src/util/element-object-generator.ts
weili-govtech, line 1: - to update with latest changes from main branch
- `generate()` and `duplicate()` have similar logic, to abstract the common parts where possible (DRY principle)

src/util/element-object-generator.ts
```diff
@@ -8,16 +8,56 @@ import { ELEMENT_BUTTON_LABELS } from "src/data";
...
+                    return generateId("numeric-field");
+                case EElementType.CONTACT:
+                    return generateId("contact-field");
+                case EElementType.TEXT:
+                    return generateId("short-text-field");
+                case EElementType.TEXTAREA:
+                    return generateId("long-text-field");
+                default:
+                    return null;
+            }
```
weili-govtech: create mapping of element type to prefix in `elements-data.ts`

weili-govtech (changes requested):

src/util/element-object-generator.ts
ShIrInRuTh: fixed in latest push

src/util/element-object-generator.ts
ShIrInRuTh: fixed in latest push

src/util/element-object-generator.ts
```diff
@@ -36,6 +74,7 @@ export namespace ElementObjectGenerator {
                     conditionalRendering: [],
                     prefill: [],
                 };
+                console.log(attributes);
```
weili-govtech: remove

weili-govtech (changes requested):

src/util/element-object-generator.ts
ShIrInRuTh: fixed in the last push

weili-govtech (approved):

PR #61, opened 2024-05-29
CCUBE-1349

ShIrInRuTh:
**Changes**

- Ensured when an element is added the form is in saved state, 
- Amended the code in the basic-details, conditional-rendering, validation & prefill to ensure that it will change the state of the form when input is changed or new entries are added.

**Additional information**

-   You may refer to this [CCUBE-1349](https://jira.ship.gov.sg/browse/CCUBE-1349)


src/components/main-panel/main-panel.tsx
```diff
@@ -78,7 +78,7 @@ export const MainPanel = () => {
     // EVENT HANDLERS
     // =========================================================================
     const handleElementCardClick = (element: TElement) => () => {
-        focusElement(element);
+        focusElement(element, false);
```
weili-govtech: now that there is no need to set the dirty state, we can remove the isDirty argument from `focusElement` and set it as false in the function

src/components/side-panel/side-panel-header.tsx
```diff
@@ -20,6 +20,7 @@ export const SidePanelHeader = () => {
         removeFocusedElement,
         focusedElement,
     } = useBuilder();
+    const isDirty = focusedElement?.isDirty || false;
```
weili-govtech:
```suggestion
   const { isDirty } = focusedElement;
```

src/components/side-panel/side-panel-header.tsx
```diff
@@ -53,9 +54,7 @@ export const SidePanelHeader = () => {
             return (
                 <>
                     <SaveChangesButton>
-                        {focusedElement.isDirty === false
-                            ? "Saved"
-                            : "Save Changes"}
+                        {isDirty === false ? "Saved" : "Save Changes"}
```
weili-govtech: instead of checking for `false`, check for `falsy` to cover undefined / null values

```suggestion
                        {!isDirty ? "Saved" : "Save Changes"}
```

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -20,10 +20,12 @@ export const BasicDetails = () => {
     const { focusedElement, updateFocusedElement } = useBuilder();
     const {
         control,
-        formState: { errors, isDirty },
+        formState: { errors, isDirty, touchedFields },
         watch,
     } = useFormContext<IBaseTextBasedFieldValues>();
     const element = focusedElement.element;
+    const shouldUpdateFocusedElement =
+        isDirty && Object.keys(touchedFields)?.length > 0;
```
weili-govtech: we shouldnt need `touchedFields`, focused element's dirty state should be dependent on RHF's dirty state only. with this, validation / conditional rendering / prefill wont need to set the dirty state individually

I also notice the form is set as dirty prematurely from the validation / conditional rendering / prefill components, can we avoid that?

weili-govtech (changes requested):

src/components/main-panel/main-panel.tsx
ShIrInRuTh: fixed in the last push

src/components/side-panel/side-panel-header.tsx
ShIrInRuTh: fixed in the last push

src/components/side-panel/side-panel-header.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/basic-details/basic-details.tsx
ShIrInRuTh: fixed in the last push

src/context-providers/builder/hook.ts
```diff
@@ -120,7 +120,7 @@ export const useBuilder = () => {
         [state.orderedIdentifiers, state.elements, state.mode]
     );
 
-    const focusElement = useCallback((element: TElement, isDirty?: boolean) => {
+    const focusElement = useCallback((element: TElement, isDirty = false) => {
```
weili-govtech: remove it as I dont see any need to set isDirty anymore
```suggestion
    const focusElement = useCallback((element: TElement) => {
```

src/components/side-panel/side-panel-header.tsx
weili-govtech: did you miss pushing this change?

src/components/side-panel/side-panel-header.tsx
weili-govtech: did you miss pushing this change?

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -38,11 +39,16 @@ export const BasicDetails = () => {
...
+        updateFocusedElement(isDirty);
     }, [isDirty, updateFocusedElement]);
 
+    useEffect(() => {
+        reset(element, {
+            keepDirty: true,
+            keepValues: true,
+        });
+    }, []);
+
```
weili-govtech, line 51: while this works, it doesnt solve the root cause of the form turning dirty prematurely - the issue is from `Validation`/`ConditionalRendering`/`Prefill` setting the RHF values manually and dirtying it

```js
useEffect(() => {
    setValue("validation", childEntryValues, {
        shouldDirty: true,
    });
}, [childEntryValues]);
```

the issue with RHF values and local states should be solved in https://github.com/LifeSG/web-form-builder/pull/58 so may I request to make sure that PR account for this too

weili-govtech (changes requested):

weili-govtech (approved):

PR #62, opened 2024-05-29
[CCUBE-1296][SR] Amended the code to ensure that the button is disabled when the input is invalid in prefill & validation.

ShIrInRuTh:
**Changes**

-  Amended the code to ensure that the button is disabled when the input is invalid in prefill & validation.

**Additional information**

-  Issue during deskcheck when showcasing was that the button was not disabled when the input was invalid.


src/components/element-editor/prefill/prefill.tsx
```diff
@@ -41,8 +41,10 @@ export const Prefill = () => {
                 });
                 return !!validationResult ? false : true;
             } catch (error) {
-                return error.errors.some((errorMessage: string | string[]) =>
-                    errorMessage.includes("required")
+                return error.errors.some(
+                    (errorMessage: string | string[]) =>
+                        errorMessage.includes("required") ||
+                        errorMessage.includes("Invalid")
                 );
```
weili-govtech: on second look, actually why do we need to check the content of the error message? shouldnt it be deemed erroneous as long as validation failed?

```suggestion
               return Yup.ValidationError.isError(error);
```

weili-govtech (changes requested):

src/components/element-editor/prefill/prefill.tsx
```diff
@@ -41,8 +41,10 @@ export const Prefill = () => {
                 });
                 return !!validationResult ? false : true;
```
weili-govtech: can deem validation successful as long as `validateSync()` does not throw error

```suggestion
                return true;
```

src/components/element-editor/prefill/prefill.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/prefill/prefill.tsx
ShIrInRuTh: fixed in the last push

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -65,11 +66,9 @@ export const ConditionalRendering = () => {
                     conditionalRendering: conditionalRenderingValues,
```
weili-govtech, line 66: no need to declare `validationResult` as it is not used

```
validationSchema.validateSync({
    prefill: prefillValues,
    abortEarly: false,
});
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #63, opened 2024-05-29
CCUBE-1351

ShIrInRuTh:
**Changes**

- Allowed toast to be shown when element is duplicated.

**Additional information**

-   You may refer to this [CCUBE-1351](https://jira.ship.gov.sg/browse/CCUBE-1351)
- There are other commits as its branched out from CCUBE-1298 to get the toast functionality to add into CCUBE-1351

### DIFF
--- a/src/components/element-card/element-card.tsx
+++ b/src/components/element-card/element-card.tsx
@@ -81,9 +81,6 @@ export const ElementCard = ({ element, onClick }: IProps) => {
             type: EToastTypes.SUCCESS_TOAST,
         };
 
-        if (disableDuplicate) {
-            event.preventDefault();
-        }
         duplicateElement(focusedElement.element);
         showToast(newToast);
     };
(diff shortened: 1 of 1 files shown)


src/components/element-card/element-card.tsx
```diff
@@ -70,11 +76,16 @@ export const ElementCard = ({ element, onClick }: IProps) => {
...
     ) => {
         event.stopPropagation();
+        const newToast = {
+            message: "Element duplicated.",
+            type: EToastTypes.SUCCESS_TOAST,
+        };
 
         if (disableDuplicate) {
             event.preventDefault();
         }
```
weili-govtech: unrelated: can remove as this will never run; user cant click it as button will be disabled

weili-govtech (changes requested):

src/components/element-card/element-card.tsx
ShIrInRuTh: fixed in latest push

weili-govtech (approved):

PR #69, opened 2024-06-12
CCUBE-1402

ShIrInRuTh:
**Changes**

- Creating search bar in add elements panel
- Enabling the search functionality
- Adding no results content when search result is not avaliable
- Added new tests for the search bar functionality.

**Additional information**

-   You may refer to this [CCUBE-1402](https://jira.ship.gov.sg/browse/CCUBE-1402)

### DIFF
--- a/src/components/side-panel/add-elements-panel/add-elements-panel.tsx
+++ b/src/components/side-panel/add-elements-panel/add-elements-panel.tsx
@@ -23,8 +23,6 @@ export const AddElementsPanel = () => {
         useState(ELEMENTS_CATEGORIES);
     const [searchInput, setSearchInput] = useState("");
 
-    const inputRef = useRef(null);
-
     // =========================================================================
     // HELPER FUNCTIONS
     // =========================================================================
@@ -52,10 +50,6 @@ export const AddElementsPanel = () => {
     // =============================================================================
     const handleElementSelectorClick = (type: EElementType) => () => {
         addElement(type, true);
-    };
-
-    const handleMagnifierClick = () => {
-        inputRef.current && inputRef.current.focus();
     };
 
     const handleOnClear = () => {
@@ -112,10 +106,7 @@ export const AddElementsPanel = () => {
                         type: "custom",
                         attributes: {
                             children: (
-                                <MagnifierIcon
-                                    onClick={handleMagnifierClick}
-                                    data-testid="maginfying-glass"
-                                />
+                                <MagnifierIcon data-testid="maginfying-glass" />
                             ),
                         },
                         position: "left",
@@ -126,7 +117,6 @@ export const AddElementsPanel = () => {
                     onChange={(event) =>
                         generateSearchResults(event.target.value)
                     }
-                    ref={inputRef}
                 />
             </label>
 
--- a/src/components/side-panel/add-elements-panel/add-elements-panel.tsx
+++ b/src/components/side-panel/add-elements-panel/add-elements-panel.tsx
@@ -1,11 +1,16 @@
 import { Text } from "@lifesg/react-design-system/text";
+import { MagnifierIcon } from "@lifesg/react-icons/magnifier";
+import { useRef, useState } from "react";
 import { ElementSelectorCard } from "src/components/element-selector-card";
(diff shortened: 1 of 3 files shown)


src/components/side-panel/add-elements-panel/add-elements-panel.tsx
```diff
@@ -51,11 +92,41 @@ export const AddElementsPanel = () => {
...
+                        ),
+                    },
+                    position: "left",
+                }}
+                allowClear={true}
+                value={searchInput}
+                onClear={handleOnClear}
+                onChange={(event) => generateSearchResults(event.target.value)}
+                ref={inputRef}
+            />
```
weili-govtech: instead of manually focusing on the search field with the `MagnifierIcon`, can we wrap it in a label instead?

```suggestion
            <label>
                <ElementSearch
                    placeholder="Search"
                    addon={{
                        type: "custom",
                        attributes: {
                            children: (
                                <MagnifierIcon
                                    onClick={handleMagnifierClick}
                                    data-testid="maginfying-glass"
                                />
                            ),
                        },
                        position: "left",
                    }}
                    allowClear={true}
                    value={searchInput}
                    onClear={handleOnClear}
                    onChange={(event) => generateSearchResults(event.target.value)}
                    ref={inputRef}
                />
            </label>
```

else you will need to use `IconWithButton` for the magnifier icon which comes with the proper accessibility tags

tests/components/add-elements-panel/add-elements-panel.spec.tsx
```diff
@@ -54,6 +54,51 @@ describe("AddElementsPanel", () => {
...
+            fireEvent.click(magnifyingGlass);
+            const searchInput = screen.getByTestId("input");
+            fireEvent.change(searchInput, { target: { value: "abc" } });
+
+            await waitFor(() => {
+                const getNoResultsFoundContent =
+                    screen.getByText("No results found.");
+                expect(getNoResultsFoundContent).toBeInTheDocument();
+            });
+        });
```
weili-govtech, line 100: add one more test to assert a category is to be hidden if it has no matching children

weili-govtech (changes requested):

src/components/side-panel/add-elements-panel/add-elements-panel.tsx
```diff
@@ -51,11 +92,45 @@ export const AddElementsPanel = () => {
...
+        <Wrapper data-testid="add-elements-panel">
+            <label>
+                <ElementSearch
+                    placeholder="Search"
+                    addon={{
+                        type: "custom",
+                        attributes: {
+                            children: (
+                                <MagnifierIcon
+                                    onClick={handleMagnifierClick}
```
weili-govtech: no need onClick anymore, remember to remove `handleMagnifierClick` function

tests/components/add-elements-panel/add-elements-panel.spec.tsx
weili-govtech: I am actually referring to this scenario where `Option group` category is hidden since it has no matching children. but I realise you may be stuck because we dont support the `Option group` yet, so may I suggest to:
- still fix this test to assert based on that scenario
- skip this test and leave a comment that it cant be run because `Option group` category is not ready yet and to unskip it when that category is added

image

weili-govtech (changes requested):

tests/components/add-elements-panel/add-elements-panel.spec.tsx
ShIrInRuTh: has been updated in the latest changes

src/components/side-panel/add-elements-panel/add-elements-panel.tsx
ShIrInRuTh: has been updated in the latest changes

tests/components/add-elements-panel/add-elements-panel.spec.tsx
weili-govtech: by skipping I meant using `it.skip()` instead of commenting the test case out

weili-govtech (approved):

PR #70, opened 2024-06-12
CCUBE-1399

ShIrInRuTh:
**Changes**

- Added error display
- Added the conditional rendering of the error display and the form builder component
- Added styles to ensure the error display looks like figma mockup
- Added tests for the functionality.

-   [keep | delete] branch

**Changelog entry**

-   [BREAKING | WARNING | remove if none] Description of change in `Component-name`

**Additional information**

-   You may refer to this [CCUBE-1399](https://jira.ship.gov.sg/browse/CCUBE-1399)

### DIFF
--- a/src/form-builder.styles.tsx
+++ b/src/form-builder.styles.tsx
@@ -1,3 +1,4 @@
+import { ErrorDisplay } from "@lifesg/react-design-system";
 import { Layout } from "@lifesg/react-design-system/layout";
 import { MediaQuery } from "@lifesg/react-design-system/media";
 import styled from "styled-components";
@@ -30,4 +31,16 @@ export const ToastWrapper = styled.div`
     flex-direction: column;
     align-items: flex-end;
     gap: 1rem;
+`;
+
+export const ErrorWrapper = styled.div`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100vh;
+
+    ${MediaQuery.MaxWidth.mobileL} {
+        width: auto;
+        padding: 0 38px 0 38px;
+    }
 `;
--- a/src/form-builder.styles.tsx
+++ b/src/form-builder.styles.tsx
@@ -2,6 +2,10 @@ import { Layout } from "@lifesg/react-design-system/layout";
 import { MediaQuery } from "@lifesg/react-design-system/media";
 import styled from "styled-components";
 
+export interface IContainerProps {
+    $isLargeScreen: boolean;
+}
+
 export const Wrapper = styled(Layout.Section)`
     height: 100%;
     overflow-x: hidden;
@@ -9,11 +13,11 @@ export const Wrapper = styled(Layout.Section)`
     padding-right: 0 !important;
 `;
 
-export const Container = styled(Layout.Container)`
+export const Container = styled(Layout.Container)<IContainerProps>`
     position: relative;
     height: 100%;
     padding: 2rem 3rem;
-
+    display: ${({ $isLargeScreen }) => ($isLargeScreen ? "block" : "none")};
     /* To accommodate to desktops of 1024px */
     ${MediaQuery.MaxWidth.tablet} {
         max-width: unset;
--- a/src/form-builder.styles.tsx
+++ b/src/form-builder.styles.tsx
@@ -17,8 +17,7 @@ export const Container = styled(Layout.Container)<IContainerProps>`
     position: relative;
     height: 100%;
     padding: 2rem 3rem;
-    display: ${({ $isLargeScreen }) => ($isLargeScreen ? "block" : "none")};
-    /* To accommodate to desktops of 1024px */
+    display: ${({ $isLargeScreen }) => ($isLargeScreen ? "grid" : "none")};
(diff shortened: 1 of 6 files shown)


src/form-builder.styles.tsx
```diff
@@ -1,3 +1,4 @@
+import { ErrorDisplay } from "@lifesg/react-design-system";
```
weili-govtech: remove unused import

src/form-builder.styles.tsx
```diff
@@ -31,3 +32,15 @@ export const ToastWrapper = styled.div`
     align-items: flex-end;
     gap: 1rem;
 `;
+
+export const ErrorWrapper = styled.div`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100vh;
```
weili-govtech: use `100%` so it can takes the parent's dimensions instead of the entire screen size because there may be other elements outside form builder - we dont want to cover those


```suggestion
    height: 100%;
```

src/form-builder.styles.tsx
```diff
@@ -31,3 +32,15 @@ export const ToastWrapper = styled.div`
...
+
+    ${MediaQuery.MaxWidth.mobileL} {
+        width: auto;
+        padding: 0 38px 0 38px;
```
weili-govtech: use rem

src/form-builder.tsx
```diff
@@ -1,23 +1,72 @@
...
+    useEffect(() => {
+        const handleResize = () => {
+            setIsLargeScreen(window.innerWidth >= 1200);
+        };
+
+        window.addEventListener("resize", handleResize);
+        return () => {
+            window.removeEventListener("resize", handleResize);
+        };
+    }, []);
```
weili-govtech:
```suggestion
    useEffect(() => {
        if (window) {
        const handleResize = () => {
            setIsLargeScreen(window.innerWidth >= 1200);
        };

        window.addEventListener("resize", handleResize);
        return () => {
            window.removeEventListener("resize", handleResize);
        };
        }
    }, []);
```

src/form-builder.tsx
```diff
@@ -1,23 +1,72 @@
...
+                <ErrorWrapper>
+                    <ErrorDisplay
+                        type="warning"
+                        title={"Screen size not supported"}
+                        description={
+                            "Expand to a bigger screen size or switch to desktop view to use the form builder."
+                        }
+                    />
+                </ErrorWrapper>
+            );
```
weili-govtech: can we move this to a new component so we can keep the logic in this file simple?

also I dont think we need a separate `renderScreen()` function since this is the only render logic here

tests/form-builder.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+        value: width,
+    });
+};
+
+describe("FormBuilder component", () => {
+    it("renders large screen content when the screen size is large", () => {
+        mockWindow(1200);
+        renderComponent();
+        expect(screen.getByText("Form is empty")).toBeInTheDocument();
+    });
```
weili-govtech: should also assert the error screen not to be rendered

weili-govtech (changes requested):

src/form-builder.tsx
```diff
@@ -23,50 +18,36 @@ export const FormBuilder = () => {
...
+                        </ToastWrapper>
+                        <MainPanel />
+                        <SidePanel />
+                    </Container>
+                </Wrapper>
+            </DisplayProvider>
+        </BuilderProvider>
+    ) : (
+        <ScreenNotSupportedErrorDisplay />
+    );
```
weili-govtech: instead of conditionally hide the form builder, can we render the `ScreenNotSupportedErrorDisplay` above it?

because the current implementation is a destructive operation - all data regardless saved or unsaved are gone since everything is removed

if this is not possible, e.g. due to z-sorting, etc, at least keep all the context providers intact even if the error is shown

weili-govtech (changes requested):

src/form-builder.tsx
ShIrInRuTh: updated in the latest changes

src/form-builder.styles.tsx
```diff
@@ -2,18 +2,22 @@ import { Layout } from "@lifesg/react-design-system/layout";
...
     padding-right: 0 !important;
 `;
 
-export const Container = styled(Layout.Container)`
+export const Container = styled(Layout.Container)<IContainerProps>`
     position: relative;
     height: 100%;
     padding: 2rem 3rem;
-
+    display: ${({ $isLargeScreen }) => ($isLargeScreen ? "block" : "none")};
```
weili-govtech:
```suggestion
    display: ${({ $isLargeScreen }) => ($isLargeScreen ? "grid" : "none")};
```

weili-govtech (changes requested):

src/form-builder.styles.tsx
ShIrInRuTh: updated in the latest changes

weili-govtech (approved):

PR #71, opened 2024-06-13
[CCUBE-1353][SR] Adding typing to the function to remove error that is blocking the merging to dev env.

ShIrInRuTh:
**Changes**
- Adding typing to the function to remove error that is blocking the merging to dev env.

### DIFF
--- a/src/components/common/toast/toasts.tsx
+++ b/src/components/common/toast/toasts.tsx
@@ -1,7 +1,7 @@
 import { useDisplay } from "../../../context-providers";
 import { DisplayToast } from "./toast";
 
-export const Toasts = () => {
+export const Toasts: Function = () => {
     const { toasts } = useDisplay();
 
     return toasts.map((toast) => (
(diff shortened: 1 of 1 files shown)


src/components/common/toast/toasts.tsx
weili-govtech, line 1: wrap in a fragment instead

```
export const Toasts = () => {
    const { toasts } = useDisplay();

    return (
        <>
            {toasts.map((toast) => (
                <DisplayToast
                    key={toast.id}
                    toast={toast}
                    onClickActionButton={toast.onClickActionButton}
                />
            ))}
        </>
    );
};
```

weili-govtech (changes requested):

src/components/common/toast/toasts.tsx
ShIrInRuTh: updated in latest changes

weili-govtech (approved):

PR #73, opened 2024-06-24
CCUBE-1431, CCUBE-1432 & CCUBE-1347

ShIrInRuTh:
**Changes**

- Added the schema generator
- Added a generateSchema() & imperative handle
- Used gridLayout for enabling single & double column layout
- Added a submit button, made function extensible
- Exposed useImperativeHandle
- Amended the form builder to allow elements to be passed into generateSchema
- Amended the generateSchema code
- Extracted code & made functions catered to generate schemas for certain field types.
- Added an offset for the formbuilder and the sidepanel to be moved down to give space for the top bar, which is exposed as a form builder prop

**Additional information**

-   You may refer to this [CCUBE-1431](https://jira.ship.gov.sg/browse/CCUBE-1431), [CCUBE-1432](https://jira.ship.gov.sg/browse/CCUBE-1432) & [CCUBE-1437](https://jira.ship.gov.sg/browse/CCUBE-1437)


package.json
```diff
@@ -36,6 +36,8 @@
         "@dnd-kit/core": "^6.0.8",
         "@dnd-kit/sortable": "^7.0.2",
         "@hookform/resolvers": "^3.3.4",
+        "@lifesg/react-design-system": "^2.6.0-canary.4",
+        "@lifesg/web-frontend-engine": "^1.10.3",
```
weili-govtech: design system and frontend engine are meant to be under `peerDependencies`, i.e. consumer is expected to have them installed in their repo rather than bundling them in this library, this keeps the library small

along with `peerDependencies`, we will install these 2 packages as `devDependencies` for dev use

src/form-builder.tsx
```diff
@@ -1,18 +1,43 @@
...
+    TElementMap,
+    useBuilder,
+} from "./context-providers/builder";
 import { Container, ToastWrapper, Wrapper } from "./form-builder.styles";
-import { ScreenNotSupportedErrorDisplay } from "./components/error-display/screen-not-supported-error";
+import { generateSchema, translateSchema } from "./util/schema-translator";
+
+export interface IFormBuilderMethods {
+    generate?: (elementsList?: TElementMap) => IFrontendEngineData;
+    translate?: (text: string) => void;
```
weili-govtech: names can be more specific
```suggestion
    generateSchema?: (elementsList?: TElementMap) => IFrontendEngineData;
    parseSchema?: (schema: string) => void;
```

src/util/schema-translator.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+};
+
+export function generateSchema(elements: TElementMap) {
+    let fields = {};
+
+    Object.entries(elements).forEach(([_, value]) => {
+        const translatedChild = generateSchemaByType(value);
+        fields = { ...fields, ...translatedChild };
+    });
```
weili-govtech: - use `Object.values()` since we are dealing with the values only
- can we use `.reduce()` instead of `.forEach()` to build the `fields` object?

src/util/schema-translator.ts
```diff
@@ -0,0 +1,92 @@
...
+import { EElementType, TElement, TElementMap } from "src/context-providers";
+
+const emailFieldToSchema = (element: TElement) => {
+    const emailFieldSchema = {
+        [element.id]: {
+            label: element.label,
+            uiType: element.type,
+            columns: {
+                desktop: element.columns,
+            },
```
weili-govtech: even though form builder is supported on desktop only, the generated form needs to be supported on all screens, do add the column size for tablet and mobile

src/util/schema-translator.ts
```diff
@@ -0,0 +1,92 @@
...
+                children: {
+                    grid: {
+                        children: {
+                            ...fields,
+                        },
+                        uiType: "grid",
+                    },
+                    "submit-button": {
+                        label: "Submit",
+                        uiType: "submit",
```
weili-govtech: can add `disabled: "invalid-form"` since we usually disable the submit button if the form is invalid

src/util/schema-translator.ts
```diff
@@ -0,0 +1,92 @@
+import { IFrontendEngineData } from "@lifesg/web-frontend-engine/components";
```
weili-govtech: FEE does not need sub-directory imports because it eagerly loads all components

```suggestion
import { IFrontendEngineData } from "@lifesg/web-frontend-engine";
```

src/util/schema-translator.ts
weili-govtech, line 1: we need more abstraction beyond this because this file will get unwieldy huge as we add the handling of other field types and the parsing from FEE schema

can we move all these logic to a new `src/translator` folder? you can follow how the `schemas/` folder is structured
- there will be a main `translator.ts` file to hold the generateSchema and parseSchema functions
- there will be a `text-based-fields.ts` to handle the generating and parsing of the text-based fields
- common logic will go into the `helper.ts` file

```
// translator.ts
export namespace Translator {
    export const generateSchema = (elements: TElementMap) => {
        // do something
    }
    export const parseSchema = (schema: string) => {
        // do something
    }
}
```

stories/doc-elements.tsx
weili-govtech, line 1: why do we need to assign a ref to the FormBuilder? how is it used?

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,79 @@
...
+    width: 100vw;
+    display: ${({ visible }) => (visible ? "block" : "none")};
+`;
+
+export const IconWrapper = styled.div`
+    display: flex;
+    width: 100%;
+    height: 5rem;
+    justify-content: flex-end;
+    z-index: 1;
```
weili-govtech, line 24: requires `position: relative` otherwise z-index wont do anything

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,79 @@
...
+    gap: 2rem;
+    padding: 1rem 1rem 0.5rem 0;
```
weili-govtech: why the uneven top and bottom padding?

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,79 @@
...
+`;
+
+export const SchemaViewWrapper = styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    gap: 2rem;
+    padding: 2rem;
+`;
```
weili-govtech: since they have the same styles, can we combine them?

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,115 @@
+import { Text } from "@lifesg/react-design-system";
```
weili-govtech:
```suggestion
import { Text } from "@lifesg/react-design-system/text";
```

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,115 @@
...
+                    onClick={handleFormPreviewButton}
+                >
+                    <EyeIcon />
+                </IconButton>
+                <IconButton
+                    $active={pageMode === "schema-mode"}
+                    onClick={handleSchemaViewButton}
+                >
+                    <TabletIcon />
+                </IconButton>
```
weili-govtech: add `type="button"` for each button, the callback wont need `e.preventDefault()`

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,115 @@
...
+            </IconWrapper>
+            <ContentWrapper visible={pageMode === "form-builder-mode"}>
+                <FormBuilder ref={formBuilderRef} offset={5.1} />
+            </ContentWrapper>
+            <ContentWrapper visible={pageMode === "preview-mode"}>
+                <FormPreview data={schema} />
+            </ContentWrapper>
```
weili-govtech: can we conditionally render the form? currently any changes / error states are retained despite switching back to the form builder and making further changes - they should be cleared

tests/util/schema-translator.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+            children: {
+                grid: {
+                    children: {
+                        ["mockElement"]: {
+                            label: "Email address",
+                            uiType: "email-field",
+                            columns: {
+                                desktop: 12,
+                            },
+                            placeholder: "",
```
weili-govtech: add a value to placeholder so we can assert it is translating correctly

tests/util/schema-translator.spec.tsx
```diff
@@ -0,0 +1,113 @@
...
+    },
+};
+
+const MOCK_TEXT_BASED_SCHEMA = {
+    sections: {
+        section: {
+            children: {
+                grid: {
+                    children: {
+                        ["mockElement"]: {
```
weili-govtech:
```suggestion
                        mockElement: {
```

weili-govtech (changes requested):

stories/doc-elements.tsx
ShIrInRuTh: without the ref, the schema is not being generated since the useImperativeHandle is in the form builder and the it relies on the elements from the form builder to run the generate schema function.

package.json
weili-govtech: the corresponding changes need to be done on the package-lock as well
the easiest way is to discard your changes to the package-lock.json and install both libraries as devdependencies again

src/util/translator/index.ts
weili-govtech, line 1: as the translator is a major feature, let's move in to `src/translator`

src/util/translator/translator.ts
```diff
@@ -0,0 +1,56 @@
...
+        };
+    }
+
+    export const generateSchema = (elements: TElementMap) => {
+        let fields = {};
+
+        fields = Object.values(elements).reduce((acc, value) => {
+            const translatedChild = elementToSchema(value);
+            return { ...acc, ...translatedChild };
+        }, {});
```
weili-govtech:
```suggestion
        const fields = Object.values(elements).reduce((acc, value) => {
            const translatedChild = elementToSchema(value);
            return { ...acc, ...translatedChild };
        }, {});
```

src/util/schema-translator.ts
weili-govtech: understand it's still WIP, just a reminder to follow what we discussed during code review

stories/doc-elements.tsx
weili-govtech: yes indeed but this `formBuilderRef` is not being used in this component right?

not to be confused with `stories/translator/doc-elements.tsx`, that is valid because the ref is used there

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    const handleFormPreviewButton = (event) => {
+        event.preventDefault();
+        setPageMode("preview-mode");
+    };
+
+    const handleSchemaViewButton = (event) => {
+        event.preventDefault();
+        setPageMode("schema-mode");
+    };
```
weili-govtech: - `event.preventDefault();` is not needed now that we added `type="button"` for each of the button
- declare typing for the event, we avoid leaving them untyped because they will default to `any`

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,73 @@
+import { Color } from "@lifesg/react-design-system";
```
weili-govtech:
```suggestion
import { Color } from "@lifesg/react-design-system/color";
```

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+import styled from "styled-components";
+
+interface IContentWrapperProps {
+    visible?: boolean;
```
weili-govtech:
```suggestion
    $visible?: boolean;
```

weili-govtech (changes requested):

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,36 @@
+import { TElement } from "src/context-providers/builder";
+
+export namespace textBasedField {
```
```suggestion
export namespace TextBasedField {
```

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,36 @@
...
+    export interface ISchemaValidation {
```
weili-govtech: no need to export
```suggestion
    interface ISchemaValidation {
```

src/context-providers/builder/element.types.ts
```diff
@@ -40,10 +40,17 @@ export interface IPrefillAttributes {
     path?: string;
 }
 
+export interface IColumns {
+    desktop: number;
+    tablet: number;
+    mobile: number;
+}
```
weili-govtech:
```suggestion
type MobileCol = 1 | 2 | 3 | 4;
type MobileColRange = MobileCol | 5;
type TabletCol = MobileCol | 5 | 6 | 7 | 8;
type TabletColRange = TabletCol | 9;
type DesktopCol = TabletCol | 9 | 10 | 11 | 12;
type DesktopColRange = DesktopCol | 13;
export interface IColumns {
    desktop: DesktopCol | [DesktopColRange, DesktopColRange];
    tablet: TabletCol | [TabletColRange, TabletColRange];
    mobile: MobileCol | [MobileColRange, MobileColRange];
}
```

src/form-builder.tsx
```diff
@@ -1,18 +1,43 @@
...
+    TElementMap,
+    useBuilder,
+} from "./context-providers/builder";
 import { Container, Wrapper } from "./form-builder.styles";
-import { ScreenNotSupportedErrorDisplay } from "./components/error-display/screen-not-supported-error";
+import { Translator } from "./translator/translator";
+
+export interface IFormBuilderMethods {
+    generateSchema?: (elementsList?: TElementMap) => IFrontendEngineData;
+    parseSchema?: (schema: string) => void;
```
weili-govtech:
```suggestion
    generateSchema: (elementsList?: TElementMap) => IFrontendEngineData;
    parseSchema: (schema: string) => void;
```

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,118 @@
...
+        setPageMode("form-builder-mode");
+    };
+
+    const handleFormPreviewButton = (event) => {
+        setPageMode("preview-mode");
+    };
+
+    const handleSchemaViewButton = (event) => {
+        setPageMode("schema-mode");
+    };
```
weili-govtech:
```suggestion
    const handleFormBuilderButton = () => {
        setPageMode("form-builder-mode");
    };

    const handleFormPreviewButton = () => {
        setPageMode("preview-mode");
    };

    const handleSchemaViewButton = () => {
        setPageMode("schema-mode");
    };
```

weili-govtech (changes requested):

PR #74, opened 2024-06-25
CCUBE-934

ShIrInRuTh:
**Changes**

- Amended the drag and drop feature according to comments given enabling 2 column drag & drop.
- Added code changes for smooth drag and dropping of elements to 2 column, abstracting of the css styles

**Additional information**

-   You may refer to this [CCUBE-934](https://jira.ship.gov.sg/browse/CCUBE-934)


src/components/main-panel/main-panel.tsx
```diff
@@ -80,37 +75,90 @@ export const MainPanel = () => {
         focusElement(element);
     };
 
-    const handleDragEnd = (event: DragEndEvent) => {
-        const { active, over } = event;
+    const handleOnDragMove = useCallback(
+        debounce((event: DragEndEvent) => {
```
weili-govtech: why do we need the debounce? without it, is it causing latency?

src/components/main-panel/main-panel.tsx
```diff
@@ -80,37 +75,90 @@ export const MainPanel = () => {
...
-            const oldIndex = orderedIdentifiers.findIndex(
+            const wrapperLeftThirds =
+                wrapperBounds.left + wrapperBounds.width / 3;
+            const wrapperRightThirds =
+                wrapperBounds.left + (wrapperBounds.width * 2) / 3;
+            const mouseX =
+                event.active.rect.current.translated.left +
+                event.activatorEvent["layerX"];
+            const { active, over } = event;
+            if (active.id === over.id) return;
```
weili-govtech: I dont see the need for this comparison actually, there are scenarios where admin wants to convert the current row into a two-column layout without other fields

src/components/main-panel/main-panel.tsx
```diff
@@ -80,37 +75,90 @@ export const MainPanel = () => {
...
+                };
+
+                const activeElement = elementIdentifiers.splice(newIndex, 1)[0];
+                elementIdentifiers.splice(oldIndex, 0, activeElement);
+                updatedOrderedIdentifiers = [...elementIdentifiers];
+            } else if (mouseX > wrapperRightThirds) {
+                // When it's dragged to the right
+                elementIdentifiers[newIndex] = {
+                    ...elementIdentifiers[newIndex],
+                    size: "third",
```
weili-govtech: I think there is a need to differentiate left vs right column, e.g. I am dragging to the right column without any fields on the left, this is not currently not possible

instead of `third`, it'll make more sense to name it `left` or `right`

src/components/element-card/element-card.styles.tsx
```diff
@@ -92,28 +94,39 @@ export const ActionButton = styled.button<IActionButtonStyleProps>`
...
-                  z-index: -1;
-              `
-            : css`
-                  border: 1px solid transparent;
-              `}
-    position: absolute;
-    flex-direction: column;
-    height: auto;
-    width: 100%;
+    position: relative;
```
weili-govtech: using relative positioning is causing extra space to appear below a dragged card, this is due to the `DroppableWrapper` taking up the space, it should be using `absolute` positioning

this does cause it to always take 100% width beyond its container though, to fix that, apply relative positioning to its direct parent

weili-govtech (changes requested):
thanks for the implementation, the basic behaviour is there already and we just need to refine further
- ability to drop to the right column without any element on the left
- ability to drop to the left / right column of the previous or next row

note: I am reviewing the drag & drop behaviour first, will review the code structure after we resolved it

src/components/main-panel/main-panel.tsx
weili-govtech: can we reduce the debounce to `10` ms?

src/components/main-panel/main-panel.tsx
```diff
@@ -80,37 +75,121 @@ export const MainPanel = () => {
...
             );
-            const updatedOrderedIdentifiers = arrayMove(
-                orderedIdentifiers,
-                oldIndex,
-                newIndex
-            );
 
+            let elementIdentifiers = [...orderedIdentifiers];
+            let updatedOrderedIdentifiers = [];
+
```
weili-govtech: thanks for the fix, I think the horizontal drag & drop generally works fine but it is becoming apparent that we need to handle the vertical drag & drop too. i.e. to determine whether admin is dragging to a new row or the current row, can we explore this?

src/components/main-panel/main-panel.styles.tsx
```diff
@@ -63,7 +63,7 @@ export const Wrapper = styled.div<IWrapperStyleProps>`
 `;
 
 export const EmptyDisplayWrapper = styled.div<IWrapperStyleProps>`
-    position: relative;
+    position: absolute;
```
weili-govtech: why does it have to be absolute positioned? this is causing the empty display to stretch to the full width, going underneath the side panel

weili-govtech (changes requested):

src/components/common/columns/columns.styles.tsx
```diff
@@ -0,0 +1,124 @@
...
+            `${MediaQuery.MaxWidth[key]} { grid-template-columns: ${value}; }`
+    )}
+`;
+
+const generateDroppableGridColumn = ($size: string) => {
+    const sizeMap = {
+        right: {
+            desktop4k: "4 / span 3",
+            desktopL: "4 / span 4",
+            tablet: "4 / span 3",
```
weili-govtech:
```suggestion
            desktopL: "5 / span 4",
            tablet: "5 / span 3",
```

src/components/main-panel/main-panel.tsx
```diff
@@ -108,38 +116,107 @@ export const MainPanel = () => {
...
+        const currentActiveIndex = orderedIdentifiers.findIndex(
+            (item) => item.internalId === active.id
+        );
+
+        const mouseX =
+            active.rect.current.translated?.left +
+            event.activatorEvent["layerX"];
+
+        let elementIndex = currentActiveIndex;
+        let column: IElementIdentifier["size"] = "full";
```
weili-govtech: let's use `size` since the rest of the logic is using that
```suggestion
        let size: IElementIdentifier["size"] = "full";
```

src/components/common/columns/columns.styles.tsx
weili-govtech, line 1: I think it's a good idea to store the various column sizes here and have a function to generate the proper styling

however the subsequent generation of the composited styling for the element card and droppable wrapper has little value; they are only used once and should be in the corresponding styled component.

to compound the issue with readability further, it is handling both max-width media query and column sizes and I am getting confused by the two sets of values. `$size` is also missing the proper union type.

can we:
- come up with a config to define the columns for left / right / full sizes, differentiated by screen sizes and expanded / minimised
- have a helper function to generate the proper styling for these sizes based on screen sizes and expanded / minimised
- revert the changes to `ElementsWrapper` styling
- update `DroppableWrapper` and `ElementItemWrapper` to use this function

weili-govtech (changes requested):

src/components/main-panel/main-panel.styles.tsx
```diff
@@ -24,38 +25,26 @@ export const Wrapper = styled.div<IWrapperStyleProps>`
...
-                `;
+                 ${MediaQuery.MaxWidth.desktop4k} {
+                    grid-column: 4 / span 6,
+                 } ${MediaQuery.MaxWidth.desktopL} {
+                    grid-column: 3 / span 8,
+                 } ${MediaQuery.MaxWidth.desktopM} {
+                    grid-column: 3 / span 8,
+                 } ${MediaQuery.MaxWidth.tablet} {
+                    grid-column: "2 / span 6,
+                 }`;
```
weili-govtech:
```
                return css`
                    ${MediaQuery.MaxWidth.desktop4k} {
                        grid-column: 4 / span 6;
                    }
                    ${MediaQuery.MaxWidth.desktopL} {
                        grid-column: 3 / span 8;
                    }
                    ${MediaQuery.MaxWidth.desktopM} {
                        grid-column: 3 / span 8;
                    }
                    ${MediaQuery.MaxWidth.tablet} {
                        grid-column: 2 / span 6;
                    }
                `;
```

src/components/common/columns/columns.styles.tsx
weili-govtech, line 1: - use this code snippet
- update `DroppableWrapper` to use `generateElementCardView()`
```
import { MediaQuery } from "@lifesg/react-design-system/media";
import { css } from "styled-components";

// CONST
const ELEMENT_CARD_SIZE_MAP = {
    expanded: {
        right: {
            desktop4k: "4 / span 3",
            desktopL: "4 / span 4",
            tablet: "4 / span 3",
        },
        left: {
            desktop4k: "1 / span 3",
            desktopL: "1 / span 4",
            tablet: "1 / span 3",
        },
        full: {
            desktop4k: "1 / span 6",
            desktopL: "1 / span 8",
            tablet: "1 / span 6",
        },
    },
    minimised: {
        right: {
            desktop4k: "4 / span 3",
            tablet: "4 / span 2",
        },
        left: {
            desktop4k: "1 / span 3",
            tablet: "1 / span 2",
        },
        full: {
            desktop4k: "1 / span 6",
            tablet: "1 / span 4",
        },
    },
};

// HELPER FUNCTIONS
const generateGridColumn = (breakpoints: { [key: string]: string }) => css`
    ${Object.entries(breakpoints).map(
        ([key, value]) =>
            `${MediaQuery.MaxWidth[key]} { grid-column: ${value}; }`
    )}
`;

export const generateElementCardView = (
    mode: "expanded" | "minimised",
    size: "left" | "right" | "full"
) => {
    return "";
    return generateGridColumn(ELEMENT_CARD_SIZE_MAP[mode][size]);
};
```

weili-govtech (changes requested):

src/components/common/columns/columns.styles.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { css } from "styled-components";
+
+// =========================================================================
+// CONST
+// =========================================================================
+const ELEMENT_CARD_SIZE_MAP = {
+    expanded: {
+        right: {
+            desktop4k: "4 / span 3",
+            desktopL: "4 / span 4",
```
weili-govtech:
```suggestion
            desktopL: "5 / span 4",
```

weili-govtech (changes requested):

src/components/main-panel/main-panel.tsx
```diff
@@ -108,38 +116,107 @@ export const MainPanel = () => {
...
+        } else if (mouseX >= wrapperRightThirds) {
+            size = "right";
+        }
+
+        if (active.rect.current.translated.top > dragStartY.current) {
+            // move down
+            elementIndex = elementStartingRefs.current.findLastIndex(
+                (elementRect) =>
+                    active.rect.current.translated.bottom + 16 >
+                    elementRect.y + elementRect.height / 2
```
weili-govtech: notice the build is failing with error for `findLastIndex`, it appears the pipelines do not support this yet so we shall use `findIndex()`

```suggestion
            elementIndex = elementStartingRefs.current.toReversed().findIndex(
                (elementRect) =>
                    active.rect.current.translated.bottom + 16 >
                    elementRect.y + elementRect.height / 2
```

src/components/main-panel/main-panel.tsx
```diff
@@ -1,26 +1,28 @@
...
     useSensors,
 } from "@dnd-kit/core";
 import {
     SortableContext,
     arrayMove,
     sortableKeyboardCoordinates,
-    verticalListSortingStrategy,
 } from "@dnd-kit/sortable";
 import { ErrorDisplay } from "@lifesg/react-design-system/error-display";
+import { debounce } from "lodash";
```
weili-govtech:
```suggestion
import debounce from "lodash/debounce";
```

weili-govtech (changes requested):

src/components/main-panel/main-panel.tsx
```diff
@@ -108,38 +116,107 @@ export const MainPanel = () => {
...
+                    active.rect.current.translated.bottom + 16 >
+                    elementRect.y + elementRect.height / 2
             );
-            const updatedOrderedIdentifiers = arrayMove(
-                orderedIdentifiers,
-                oldIndex,
-                newIndex
+        } else {
+            // move up
+            elementIndex = [...elementStartingRefs.current.reverse()].findIndex(
```
weili-govtech: apologies, wrong brackets

```suggestion
            elementIndex = [...elementStartingRefs.current].reverse().findIndex(
```

weili-govtech:
resolve merge conflicts

src/components/main-panel/main-panel.tsx
```diff
@@ -103,38 +116,111 @@ export const MainPanel = () => {
...
+        } else {
+            // move up
+            elementIndex = [...elementStartingRefs.current]
+                .reverse()
+                .findIndex(
+                    (elementRect) =>
+                        active.rect.current.translated.bottom + 16 >
+                        elementRect.y + elementRect.height / 2
+                );
+        }
```
weili-govtech:
```suggestion
       if (active.rect.current.translated.top > dragStartY.current) {
            // move down
            elementIndex = elementStartingRefs.current.findIndex(
                (elementRect) =>
                    active.rect.current.translated.bottom + 16 >
                    elementRect.y + elementRect.height / 2
            );
        } else {
            // move up
            const reversedIndex = [...elementStartingRefs.current]
                .reverse()
                .findIndex(
                    (elementRect) =>
                        active.rect.current.translated.bottom + 16 >
                        elementRect.y + elementRect.height / 2
                );
            const elementCount = elementStartingRefs.current.length - 1;
            elementIndex = elementCount - reversedIndex;
        }
```

weili-govtech (approved):

PR #75, opened 2024-06-26
CCUBE-1433

adenteo:
**Changes**
- Made a file to keep text-based field schema generation in.
- Amending test based on the changes made for the translator functionality.
- Made changes for function that creates validation schema, making it extensible.
- Rearranged the translator codes accordingly.
- Amended the tests accordingly.
- Added validation & conditional rendering schema generation fields.
- Added tests to ensure generation is working as intended.
- Added enums for conditional rendering comparator.
- Made tests for the generation of schema for both email and text based fields.
- Added a new component to test the form & schema preview.
- Amended the form builder to allow elements to be passed into generateSchema
- Amended the generateSchema code, extracted code.
- Made functions catered to generate schemas for certain field types.
- Added the schema generator, added a generateSchema() imperative handle.
- Used gridLayout for enabling single & double column layout.
- Added a submit button.
- Made function extensible.
- Exposed useImperativeHandle.
- Created stories to test the generation of the schema & the preview modes.

**Additional information**
-   You may refer to this [CCUBE-1433](https://jira.ship.gov.sg/browse/CCUBE-1433)


src/util/translator/helper.ts
weili-govtech, line 1: instead of `.map()`, you should use `.reduce()` or `.forEach()` because there can be multiple conditional rendering rules for the same source field, e.g. show field2 if field1 is equals to "a" or "b"

weili-govtech (changes requested):

PR #76, opened 2024-06-27
CCUBE-1431, CCUBE-1432, CCUBE-1433, CCUBE-1434 & CCUBE-1437

adenteo:
**Changes**
- Add schema generation, unit testing and amended typings for schema generation for prefill.

**Additional information**
-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1434)


src/context-providers/builder/element.types.ts
```diff
@@ -40,10 +47,23 @@ export interface IPrefillAttributes {
...
+    desktop: DesktopCol | [DesktopColRange, DesktopColRange];
+    tablet: TabletCol | [TabletColRange, TabletColRange];
+    mobile: MobileCol | [MobileColRange, MobileColRange];
+}
+
 export interface IBaseFieldAttributes extends IBaseAttributes {
     label: string | undefined;
     required: boolean;
     requiredErrorMsg?: string | undefined;
+    columns?: IColumns;
```
weili-govtech: since we can be dealing with two-column layouts, it is prudent to ensure the columns are set all times

```suggestion
    columns: IColumns;
```

src/form-builder.tsx
```diff
@@ -1,18 +1,52 @@
-import { useEffect, useState } from "react";
+import { IFrontendEngineData } from "@lifesg/web-frontend-engine/components/types";
```
weili-govtech: FEE doesnt need sub-directory because it has to load all its components to determine what components to use - there is no savings unlike DS

```suggestion
import { IFrontendEngineData } from "@lifesg/web-frontend-engine";
```

src/form-builder.tsx
```diff
@@ -34,22 +68,26 @@ export const FormBuilder = () => {
...
+    ({ offset }, ref) => {
+        return (
+            <BuilderProvider>
+                <DisplayProvider>
+                    <Component ref={ref} offset={offset} />
+                </DisplayProvider>
+            </BuilderProvider>
+        );
+    }
+);
```
weili-govtech: there is no point destructuring the props prematurely - they arent use here and new props have to be added manually

```suggestion
export const FormBuilder = forwardRef<IFormBuilderMethods, IProps>(
    (props, ref) => {
        return (
            <BuilderProvider>
                <DisplayProvider>
                    <Component ref={ref} {...props} />
                </DisplayProvider>
            </BuilderProvider>
        );
    }
);
```

src/translator/helper.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (value.prefill) {
+            return {
+                [value.id]: value.prefill,
+            };
+        }
+    });
+    Object.values(prefillObj).forEach((prefilObjChild) => {
+        prefill = { ...prefill, ...prefilObjChild };
+    });
+    return prefill;
```
weili-govtech: - combine both iterators into a single `Object.values(elements).reduce()`
- can we store the prefill settings in an array instead, it's possible for multiple prefill configs for a single field (sorry I missed that out), so the shape will be

```
{
    prefill: {
        [fieldId: string]: [
            {
                mode: "Myinfo" | "Previous source";
                actionId: undefined | string;
                path: string;
            }
        ]
    }
}
```

src/translator/helper.ts
```diff
@@ -0,0 +1,53 @@
...
+                    },
+                    {
+                        [SCHEMA_CONDITION_TYPES[condition.comparator]]: value,
+                    },
+                ];
+            } else {
+                acc[condition.fieldKey].push({
+                    [SCHEMA_CONDITION_TYPES[condition.comparator]]: value,
+                });
+            }
```
weili-govtech: - why do we need the `filled` rule?
- can move the array push out of the if...else block to avoid repeating codes

```suggestion
            if (!acc[condition.fieldKey]) {
                acc[condition.fieldKey] = [];
            }
            acc[condition.fieldKey].push({
                [SCHEMA_CONDITION_TYPES[condition.comparator]]: value,
            });
```

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,89 @@
+import { EElementType, IValidation, TElement } from "src/context-providers";
+import { createConditionalRenderingObject } from "./helper";
+
+export namespace textBasedField {
+    export namespace email {
+        export interface ISchemaValidation {
+            [key: string]: string | boolean;
+            errorMessage: string;
+        }
```
weili-govtech: is `ISchemaValidation` intended to belong to the `email` namespace only or to be shared among all text-based fields?

if it's the former, it should be outside the email scope and does not need to be exported.

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,89 @@
+import { EElementType, IValidation, TElement } from "src/context-providers";
+import { createConditionalRenderingObject } from "./helper";
+
+export namespace textBasedField {
```
```suggestion
export namespace TextBasedField {
```

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,89 @@
...
+    export namespace email {
```
```suggestion
    namespace Email {
```

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,89 @@
...
+            };
+
+            return {
+                matches: domainRegexString(validation).toString(),
+                errorMessage: validation.validationErrorMessage,
+            };
+        };
+    }
+
+    export const createValidationObject = (element: TElement) => {
```
```suggestion
    const createValidationObject = (element: TElement) => {
```

src/translator/text-based-field.ts
```diff
@@ -0,0 +1,89 @@
...
+        const validation: email.ISchemaValidation[] = [
+            {
+                required: element.required,
+                errorMessage: element.requiredErrorMsg,
+            },
+        ];
```
weili-govtech: can we assign the required validation only if the required rule is true? otherwise the generated schema will always contain the required rule

src/translator/translator.ts
```diff
@@ -0,0 +1,43 @@
...
+import { textBasedField } from "./text-based-field";
+import { createPrefillObject } from "./helper";
+
+export namespace Translator {
+    export function generateSchema(elements: TElementMap) {
+        let fields = {};
+
+        const prefill = createPrefillObject(elements);
+
+        Object.entries(elements).forEach(([_, value]) => {
```
weili-govtech: - use `Object.values()` since the key is not needed
- use `.reduce()` to build the fields

src/translator/translator.ts
```diff
@@ -0,0 +1,43 @@
...
+                            disabled: "invalid-form",
+                            label: "Submit",
+                            uiType: "submit",
+                        },
+                    },
+                    uiType: "section",
+                },
+            },
+        };
+        return { schema: elementSchema, prefill: prefill };
```
weili-govtech:
```suggestion
        return { schema: elementSchema, prefill };
```

src/translator/translator.ts
```diff
@@ -0,0 +1,43 @@
+import { IFrontendEngineData } from "@lifesg/web-frontend-engine/components";
```
weili-govtech:
```suggestion
import { IFrontendEngineData } from "@lifesg/web-frontend-engine";
```

src/util/element-object-generator.ts
```diff
@@ -70,6 +70,7 @@ export namespace ElementObjectGenerator {
                     required: true,
                     requiredErrorMsg: "",
                     label: ELEMENT_BUTTON_LABELS[type],
+                    columns: { desktop: 12, tablet: 12, mobile: 4 },
```
weili-govtech:
```suggestion
                    columns: { desktop: 12, tablet: 8, mobile: 4 },
```

stories/translator/doc-elements.tsx
```diff
@@ -0,0 +1,118 @@
+import { Text } from "@lifesg/react-design-system/text";
+import { EyeIcon, PencilIcon, TabletIcon } from "@lifesg/react-icons";
+import {
+    FrontendEngine,
+    IFrontendEngineData,
+} from "@lifesg/web-frontend-engine";
```
weili-govtech:
```suggestion
import { EyeIcon } from "@lifesg/react-icons/eye";
import { PencilIcon } from "@lifesg/react-icons/pencil";
import { TabletIcon } from "@lifesg/react-icons/tablet";
import { FrontendEngine } from "@lifesg/web-frontend-engine";
```

stories/translator/doc-elements.styles.tsx
```diff
@@ -0,0 +1,73 @@
...
+    padding: 2rem;
+`;
+
+export const SchemaPreview = styled.div`
+    margin: 2rem;
+    width: 80vw;
+    max-height: 31.25rem;
+    padding: 1rem;
+    overflow: auto;
+    border: 0.063rem solid #ccc;
```
weili-govtech: - use px for border since we wont scale it according to the font size
- use the same border colour as the fields
```suggestion
    border: 1px solid ${Color.Neutral[5]};
```

tests/util/translator/translator.spec.tsx
```diff
@@ -0,0 +1,498 @@
...
+                        disabled: "invalid-form",
+                        label: "Submit",
+                        uiType: "submit",
+                    },
+                },
+                uiType: "section",
+            },
+        },
+    },
+};
```
weili-govtech: instead of manually creating a mock for each test case, can we have a function generate that instead?

it can come with standard properties and allow overriding in areas where needed.

also, avoid the square brackets for keys as that is for dynamic keys only

weili-govtech (changes requested):

tests/util/translator/translator.spec.tsx
```diff
@@ -0,0 +1,311 @@
...
+            uiType: EElementType.EMAIL,
+            validation: [
+                {
+                    required: true,
+                    errorMessage: "Email address is required",
+                },
+            ],
+        }),
+    }
+);
```
weili-govtech: since all the mock elements and schemas are used only once, you can declare them within the test case using them. we want to declare outside only if they are shared among multiple test cases

tests/util/translator/translator.spec.tsx
```diff
@@ -0,0 +1,311 @@
...
+            disconnect: jest.fn(),
+        }));
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+        jest.resetAllMocks();
+    });
+    describe("generateSchema", () => {
+        describe('Email field schema generation"', () => {
```
weili-govtech: - use double quotes for test description
- errant double quote at the end of the description
- do this for all the test descriptions in this file
```suggestion
        describe("Email field schema generation", () => {
```

tests/util/translator/helper.ts
```diff
@@ -0,0 +1,131 @@
...
+            requiredErrorMsg: requiredErrorMsg || "",
+            columns: columns || { desktop: 12, tablet: 8, mobile: 4 },
+            placeholder: placeholder || "",
+            label: label || ELEMENT_BUTTON_LABELS[type],
+            validation: validation || [],
+            conditionalRendering: conditionalRendering || [],
+            prefill: prefill || [],
+        },
+    };
+};
```
weili-govtech: this can be simplified

```suggestion
export const generateMockElement = (props: MockElementProps): TElementMap => {
    return {
        [props.id]: {
            required: true,
            columns: { desktop: 12, tablet: 8, mobile: 4 },
            label: ELEMENT_BUTTON_LABELS[props.type],
            ...props,
        },
    };
};
```

tests/util/translator/helper.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export const generateMockSchema = (
+    prefill: MockPrefillSchema,
+    children: MockChildrenSchema
+) => {
+    return {
+        prefill: prefill,
```
weili-govtech:
```suggestion
        prefill,
```

tests/util/translator/helper.ts
```diff
@@ -0,0 +1,131 @@
...
+    label,
+    uiType,
+    placeholder,
+    validation,
+    showIf,
+}: MockElementSchemaProps): MockChildrenSchema => {
+    return {
+        [id]: {
+            label: label,
+            uiType: uiType,
```
weili-govtech:
```suggestion
            label,
            uiType,
```

tests/util/translator/translator.spec.tsx
```diff
@@ -70,40 +75,33 @@ describe("Translator", () => {
...
 // =============================================================================
 
+const MOCK_TEXT_BASED_ELEMENT: TElementMap = {
+    ...generateMockElement({
+        type: EElementType.TEXT,
+        id: "mockId1",
+        internalId: "mockInternalId1",
+        requiredErrorMsg: "Input is required",
+    }),
+};
```
weili-govtech: no need to spread since the function is already generating a new object
we typically spread to do shallow cloning (you can read up more if you are unsure)

```suggestion
const MOCK_TEXT_BASED_ELEMENT: TElementMap = generateMockElement({
        type: EElementType.TEXT,
        id: "mockId1",
        internalId: "mockInternalId1",
        requiredErrorMsg: "Input is required",
    });
```

weili-govtech (changes requested):

tests/util/translator/helper.ts
```diff
@@ -0,0 +1,112 @@
+import {
+    EElementType,
+    IColumns,
+    IConditionalRendering,
+    IPrefillAttributes,
+    IValidation,
+    TElement,
```
weili-govtech: remove unused import

weili-govtech (changes requested):

weili-govtech (approved):

PR #77, opened 2024-07-02
CCUBE-1438

adenteo:
**Changes**
- Add onSubmit prop to FormBuilder
- Add action/state to toggle/show submitting state in builder context.
- Edit save button to show disabled loading state based on submitting state.
- Add On Submit story to main stories.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1438)

### DIFF
--- a/src/context-providers/builder/provider.tsx
+++ b/src/context-providers/builder/provider.tsx
@@ -17,7 +17,6 @@ const DEFAULT_VALUES: IBuilderState = {
     focusedElement: null,
     showSidePanel: true,
     orderedIdentifiers: [],
-    isSubmitting: false,
 };
 
 // =============================================================================
@@ -28,10 +27,6 @@ export const builderReducer = (
     action: TBuilderAction
 ) => {
     switch (action.type) {
-        case "toggle-submitting": {
-            state.isSubmitting = action.payload;
-            break;
-        }
         case "toggle-panel": {
             state.showSidePanel = action.payload;
             break;
--- a/src/context-providers/builder/provider.tsx
+++ b/src/context-providers/builder/provider.tsx
@@ -17,6 +17,7 @@ const DEFAULT_VALUES: IBuilderState = {
     focusedElement: null,
     showSidePanel: true,
     orderedIdentifiers: [],
+    isSubmitting: false,
 };
 
 // =============================================================================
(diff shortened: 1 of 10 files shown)


src/context-providers/builder/provider.tsx
```diff
@@ -17,6 +17,7 @@ const DEFAULT_VALUES: IBuilderState = {
     focusedElement: null,
     showSidePanel: true,
     orderedIdentifiers: [],
+    isSubmitting: false,
```
weili-govtech: can we track it within `SidePanel` as a local state instead? because I see it being used in `SidePanelHeader` only

stories/main.stories.tsx
```diff
@@ -1,12 +1,28 @@
...
     component: DocElement,
+    /**
+     * Below is a workaround for a type error in the storybook addon when using an async function.
+     * https://github.com/storybookjs/storybook/issues/23731
+     */
+    args: {
+        onSubmit: async (...args) => {
+            action("onSubmit")(args);
+        },
+    },
```
weili-govtech: I dont see any type errors even after commenting this block of code, is it still present?

weili-govtech (changes requested):
a few observations
- where possible, keep the PR small and concise, exclude commits from other un-merged branches, this means changes from CCUBE-1434 should not be in this PR
- I realise we neglected disabling the form while it is being submitted, can we explore using the `inert` prop to disable the entire form? can read up on this thread regarding type error too: https://stackoverflow.com/questions/72720469/error-when-using-inert-attribute-with-typescript
- can we have unit tests to cover the submit behaviour?

src/context-providers/builder/types.ts
```diff
@@ -52,6 +52,11 @@ export interface IBuilderState {
 // =============================================================================
 // ACTIONS
 // =============================================================================
+export interface IToggleSubmittingAction {
+    type: "toggle-submitting";
+    payload: boolean;
+}
+
```
weili-govtech: now that we have moved `isSubmitting` to side panel, this action is not needed anymore and should be removed

tests/components/side-panel/side-panel.spec.tsx
```diff
@@ -12,11 +13,14 @@ describe("SidePanel", () => {
             unobserve: jest.fn(),
             disconnect: jest.fn(),
         }));
+        jest.useFakeTimers();
```
weili-govtech: by mocking the timers, `waitFor` will actually fail to work properly, hence we should:
- remove mocking the timers
- `mockOnSubmit` will still continue to be mocked with setTimeout, just without a delay so it will be resolved in the next hoop

weili-govtech (changes requested):

weili-govtech (approved):

PR #78, opened 2024-07-02
CCUBE-1435

ShIrInRuTh:
**Changes**
- Added a translate schema function
- Added a new hook to update elements and added the hook to the useImperative handler
- Added new typings for the hook
- Added new enums for conditional rendering translation
- Added new function to refactor the elements to the correct form before updating form builder elements state
- Removed the returning of elemenst when running the parse schema function
- Made unit tests to test the new functionality.

**Additional information**
-   You may refer to this [CCUBE-1435](https://jira.ship.gov.sg/browse/CCUBE-1435)

### DIFF
--- a/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
+++ b/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
@@ -30,7 +30,6 @@ export const ConditionalRendering = () => {
         useFormContext<IBaseTextBasedFieldValues>();
 
     const schema = SchemaHelper.buildSchema(EElementType.EMAIL);
-    const invalidAndEmptyFields = checkIsValid();
     const conditionalRenderingValues = getValues("conditionalRendering") || [];
     // =====================================================================
     // HELPER FUNCTIONS
@@ -66,6 +65,7 @@ export const ConditionalRendering = () => {
     }
 
     const getPopoverMessage = useCallback(() => {
+        const invalidAndEmptyFields = checkIsValid();
         if (invalidAndEmptyFields) {
             return (
                 <Text.Body>
@@ -77,7 +77,7 @@ export const ConditionalRendering = () => {
             return <Text.Body>No conditional rendering available.</Text.Body>;
         }
         return null;
-    }, [invalidAndEmptyFields, getElementOptions]);
+    }, [checkIsValid(), getElementOptions]);
 
     // =============================================================================
     // EVENT HANDLERS
@@ -148,9 +148,7 @@ export const ConditionalRendering = () => {
             onAdd={handleAddButtonClick}
             title="Conditional Rendering"
             buttonLabel="condition"
-            disabledButton={
-                Object.values(elements).length === 1 || invalidAndEmptyFields
-            }
+            disabledButton={getElementOptions().length === 0 || checkIsValid()}
             popoverMessage={getPopoverMessage()}
         >
             {renderChildren()}
--- a/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
+++ b/src/components/element-editor/conditional-rendering/conditional-rendering.tsx
@@ -149,7 +149,7 @@ export const ConditionalRendering = () => {
             title="Conditional Rendering"
             buttonLabel="condition"
             disabledButton={
-                getElementOptions().length === 0 || invalidAndEmptyFields
+                Object.values(elements).length === 1 || invalidAndEmptyFields
             }
             popoverMessage={getPopoverMessage()}
         >
--- a/src/context-providers/builder/hook.ts
+++ b/src/context-providers/builder/hook.ts
@@ -171,22 +171,13 @@ export const useBuilder = () => {
 
     const updateElementSchema = useCallback(
         (elements?: TElementMap, orderedIdentifiers?: IElementIdentifier[]) => {
-            if (orderedIdentifiers) {
-                dispatch({
-                    type: "update-schema-element",
-                    payload: {
-                        elements: elements,
-                        orderedIdentifiers: orderedIdentifiers,
-                    },
-                });
-            } else {
-                dispatch({
-                    type: "update-schema-element",
-                    payload: {
-                        elements: elements,
-                    },
-                });
-            }
+            dispatch({
+                type: "update-schema-element",
+                payload: {
+                    elements,
+                    orderedIdentifiers,
+                },
+            });
         },
         []
     );
--- a/src/context-providers/builder/hook.ts
+++ b/src/context-providers/builder/hook.ts
@@ -115,7 +115,7 @@ export const useBuilder = () => {
                         {
                             ...child,
                             conditionalRendering:
-                                child.conditionalRendering.filter(
+                                child?.conditionalRendering?.filter(
                                     (condition) =>
                                         condition.internalId !== internalId
                                 ),
@@ -169,6 +169,28 @@ export const useBuilder = () => {
         });
     }, []);
 
+    const updateElementSchema = useCallback(
+        (elements?: TElementMap, orderedIdentifiers?: IElementIdentifier[]) => {
+            if (orderedIdentifiers) {
+                dispatch({
+                    type: "update-schema-element",
+                    payload: {
+                        elements: elements,
+                        orderedIdentifiers: orderedIdentifiers,
+                    },
+                });
+            } else {
+                dispatch({
+                    type: "update-schema-element",
+                    payload: {
+                        elements: elements,
+                    },
+                });
+            }
+        },
+        []
+    );
+
     const updateFocusedElement = useCallback(
         (isDirty: boolean, element?: TElement) => {
             dispatch({
@@ -198,5 +220,6 @@ export const useBuilder = () => {
         updateElement,
         updateFocusedElement,
         duplicateElement,
+        updateElementSchema,
     };
 };
--- a/src/translator/helper.ts
+++ b/src/translator/helper.ts
@@ -1,9 +1,19 @@
 import {
     EConditionType,
     IConditionalRendering,
+    TElement,
     TElementMap,
 } from "src/context-providers/builder";
-import { SCHEMA_CONDITION_TYPES } from "src/data";
+import { ELEMENT_CONDITION_TYPES, SCHEMA_CONDITION_TYPES } from "src/data";
+import { generateNewInternalId } from "src/util";
+
+interface ISchemaConditionChild {
+    [comparator: string]: string | boolean;
+}
+
+interface ISchemaCondition {
+    [key: string]: ISchemaConditionChild[];
+}
 
 export interface ISchemaConditionalRendering {
     [key: string]: {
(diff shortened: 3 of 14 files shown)


src/translator/translator.ts
```diff
@@ -35,7 +36,12 @@ export namespace Translator {
...
-        console.log("this is the schema:", schema);
-    }
+    export const parseSchema = (schema: ISchemaProps) => {
+        const schemaToTranslate = (schema?.schema as any)?.sections?.section
+            ?.children?.grid?.children as any;
+        const translatedElements =
+            TextBasedField.translateToElement(schemaToTranslate);
+
+        return updateTranslatedElements(translatedElements);
+    };
```
weili-govtech: - use different name to improve readability
- unnecessary castings
- avoid casting as any

```suggestion
    export const parseSchema = (formSchema: ISchemaProps) => {
       const schemaToTranslate = formSchema?.schema?.sections?.section
            ?.children?.grid?.["children"] as Record<
            string,
            TFrontendEngineFieldSchema
        >;
        const translatedElements =
            TextBasedField.translateToElement(schemaToTranslate);

        return updateTranslatedElements(translatedElements);
    };
```

src/translator/text-based-field.ts
```diff
@@ -95,4 +139,36 @@ export namespace TextBasedField {
 
         return textBasedFieldSchema;
     };
+
+    export const translateToElement = (schemaElements: {
+        [key: string]: any;
+    }) => {
```
weili-govtech: - shouldnt this function translate 1 element instead of looping through the entire schema? because when we have other type of fields, e.g. options-based fields that dont belong here, we will be looping multiple times. instead of this, the loop should take place in `parseSchema` and call this function only if it is a text-based field
- can rename the function to `schemaToElement`, so we know it is the opposite of `elementToSchema`

src/translator/text-based-field.ts
```diff
@@ -95,4 +139,36 @@ export namespace TextBasedField {
...
+        const translatedElements: TElement[] = [];
+
+        Object.entries(schemaElements).forEach(([key, element]) => {
+            const { showIf, uiType, validation, ...rest } = element;
+            const remainingValidation =
+                validation.length > 1 ? validation.slice(1) : [];
+
+            translatedElements.push({
```
weili-govtech: can assign `internalId` here so you dont need to update elsewhere

src/translator/text-based-field.ts
```diff
@@ -95,4 +139,36 @@ export namespace TextBasedField {
...
+
+    export const translateToElement = (schemaElements: {
+        [key: string]: any;
+    }) => {
+        const translatedElements: TElement[] = [];
+
+        Object.entries(schemaElements).forEach(([key, element]) => {
+            const { showIf, uiType, validation, ...rest } = element;
+            const remainingValidation =
+                validation.length > 1 ? validation.slice(1) : [];
```
weili-govtech: other elements can support multiple validations so we shouldnt be picking just 1 validation only, instead we should be ensuring the validation configs passed to it are supported

src/translator/helper.ts
```diff
@@ -52,3 +62,69 @@ export const createConditionalRenderingObject = (
...
+                        comparator: ELEMENT_CONDITION_TYPES[comparator],
+                        value: compValue,
+                    });
+                });
+        });
+        return translatedConditions;
+    }, []);
+};
+
+export const updateTranslatedElements = (schemaElements: TElement[]) => {
```
weili-govtech: it will be better to refine this function to just handle building the conditional rendering properties only, this way we can keep the function small

for `internalId`, it should be created as part of each element 

for ordered identifiers, we can move it to `parseSchema` as that should be iterating through the schema

stories/translator/doc-elements.styles.tsx
weili-govtech, line 1: observed horizontal scrolling in the story

stories/translator/doc-elements.tsx
```diff
@@ -4,19 +4,28 @@ import { PencilIcon } from "@lifesg/react-icons/pencil";
...
+    ViewWrapper,
 } from "./doc-elements.styles";
 
 interface IProps {
     data?: ISchemaProps;
 }
 
+interface ISchemaViewProps {
+    data?: ISchemaProps;
+    setSchemaFromView?: (schema: ISchemaProps) => void;
```
weili-govtech: rename as `onChange`

stories/translator/doc-elements.tsx
```diff
@@ -26,13 +35,35 @@ const FormPreview = ({ data }: IProps) => {
     );
 };
 
-const SchemaView = ({ data }: IProps) => {
+const SchemaView = ({ data, setSchemaFromView }: ISchemaViewProps) => {
+    const [schema, setSchema] = useState("");
+
+    const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+        setSchema(event.target.value);
+    };
```
weili-govtech: is it necessary to set the schema on change? will it work to set in the blur event?

weili-govtech (changes requested):

src/translator/text-based-field.ts
ShIrInRuTh: for this it removes the required validation object, like in the img, the remaining validation is being assigned as remainingValidation.
image

weili-govtech: but the required validation may not be the first entry

src/translator/text-based-field.ts
```diff
@@ -95,4 +140,36 @@ export namespace TextBasedField {
 
         return textBasedFieldSchema;
     };
+
+    export const translateToElement = (element, key: string) => {
+        const { showIf, uiType, validation, ...rest } = element;
+        const requiredValidation = validation.filter(
+            (child: { hasOwnProperty: (arg0: string) => any }) =>
+                !Object.prototype.hasOwnProperty.call(child, "required")
+        );
```
weili-govtech: `requiredValidation` is not an accurate name because it is checking for non-required rules, moreover, you need to check for the existence for the required rule.

the current way of doing `required: validation?.[0]?.required` is not sufficient because the rule may not always be at the first entry.

so there are two things you want to achieve here:
1. check for the existence of required rule and ensuring it is true to pass to `required` property
2. find the remaining rules to pass to the `validation` array

instead of doing a filter, you can use `forEach` to construct the 2 variables.

stories/translator/doc-elements.styles.tsx
weili-govtech: this is not solved?

image

src/translator/text-based-field.ts
```diff
@@ -95,4 +140,36 @@ export namespace TextBasedField {
 
         return textBasedFieldSchema;
     };
+
+    export const translateToElement = (element, key: string) => {
```
weili-govtech: missing typing for `element`

src/translator/helper.ts
```diff
@@ -52,3 +64,66 @@ export const createConditionalRenderingObject = (
...
+                        value: compValue,
+                        internalId,
+                    });
+                });
+        });
+        return translatedConditions;
+    }, []);
+};
+
+export const translateSchemaBasedOnType = (
```
weili-govtech: since we are using `parseSchema` in the imperative handle, let's keep the naming consistent, replace `translateSchema` prefixes with `parseSchema`

src/translator/helper.ts
```diff
@@ -52,3 +64,66 @@ export const createConditionalRenderingObject = (
...
+            case EElementType.TEXT:
+            case EElementType.TEXTAREA: {
+                translatedElements.push(
+                    TextBasedField.translateToElement(element, key)
+                );
+                break;
+            }
+            default: {
+                return;
+            }
```
weili-govtech: not necessary, can remove

weili-govtech (changes requested):

src/translator/text-based-field.ts
```diff
@@ -1,12 +1,35 @@
...
+    EElementType,
+    IColumns,
+    IValidation,
+    TElement,
+} from "src/context-providers";
+import { ELEMENT_VALIDATION_TYPES } from "src/data";
+import { SimpleIdGenerator } from "src/util/simple-id-generator";
+import {
+    ISchemaCondition,
+    ISchemaConditionalRendering,
```
weili-govtech: remove unused import

src/translator/helper.ts
weili-govtech: can we update the rest of the names?
- updateTranslatedElements
- translateToElement
- translateConditionalRenderingObject
- translatedConditions
- schemaToTranslate
- translatedElements

src/context-providers/builder/hook.ts
```diff
@@ -169,6 +169,28 @@ export const useBuilder = () => {
...
+                    },
+                });
+            } else {
+                dispatch({
+                    type: "update-schema-element",
+                    payload: {
+                        elements: elements,
+                    },
+                });
+            }
```
weili-govtech: - is it necessary to exclude `orderedIdentifiers` if it is not defined?
- use shorthand if key and variable name are the same

stories/translator/doc-elements.tsx
```diff
@@ -64,9 +88,11 @@ export const DocElement = () => {
...
     // =========================================================================
     useEffect(() => {
-        if (formBuilderRef.current) {
+        if (formBuilderRef.current && pageMode !== "form-builder-mode") {
             const generatedSchema = formBuilderRef.current?.generateSchema();
             setSchema(generatedSchema);
+        } else if (formBuilderRef.current && schema) {
+            formBuilderRef.current?.parseSchema(schema);
         }
     }, [formBuilderRef.current, pageMode]);
```
weili-govtech: actually ref do not cause a rerender so adding it as a dependency is of no use

src/translator/helper.ts
```diff
@@ -52,3 +64,66 @@ export const createConditionalRenderingObject = (
 
     return Object.keys(conditionObj).length === 0 ? [] : [conditionObj];
 };
+
+export const translateConditionalRenderingObject = (
+    conditions: ISchemaCondition[],
```
weili-govtech: where possible, use the typings from FEE since it is meant to parse from it
```suggestion
    conditions: IYupConditionalValidationRule[],
```

src/translator/text-based-field.ts
```diff
@@ -1,12 +1,35 @@
...
+    export interface IElementSchema {
+        columns: IColumns;
+        label: string;
+        placeholder: string;
+        required: boolean;
+        requiredErrorMessage: string;
+        uiType: EElementType;
+        validation?: ISchemaValidation[];
+        showIf?: ISchemaCondition[];
+    }
```
weili-govtech: this interface is not necessary as we should use FEE schema where possible, can replace with `ITextFieldSchema | IEmailFieldSchema | INumericFieldSchema`

src/translator/text-based-field.ts
```diff
@@ -27,11 +50,34 @@ export namespace TextBasedField {
                 errorMessage: validation.validationErrorMessage,
             };
         };
+
+        export const translateEmailValidation = (
+            validation: ISchemaValidation[]
```
```suggestion
            validation: IYupValidationRule[]
```

src/translator/text-based-field.ts
```diff
@@ -68,6 +114,23 @@ export namespace TextBasedField {
         return validation;
     };
 
+    export const translateValidationObject = (
+        type: EElementType,
+        validation: ISchemaValidation[]
```
```suggestion
        validation: IYupValidationRule[]
```

weili-govtech (changes requested):
observed a few things
- for conditional rendering, the `Add condition` button is no longer disabled after parsing the schema
- since we are parsing FEE schema, we should base the typing on FEE schema where possible, avoid creating own interfaces because they will not match FEE

weili-govtech:
> for conditional rendering, the Add condition button is no longer disabled after parsing the schema

still observed this issue

src/translator/text-based-field.ts
```diff
@@ -27,12 +39,35 @@ export namespace TextBasedField {
                 errorMessage: validation.validationErrorMessage,
             };
         };
+
+        export const translateEmailValidation = (
```
```suggestion
        export const parseEmailValidation = (
```

src/translator/text-based-field.ts
```diff
@@ -68,6 +103,23 @@ export namespace TextBasedField {
         return validation;
     };
 
+    export const translateValidationObject = (
```
```suggestion
    export const parseValidationObject = (
```

src/translator/text-based-field.ts
```diff
@@ -95,4 +147,43 @@ export namespace TextBasedField {
...
+        validation.forEach((rule) => {
+            if (Object.prototype.hasOwnProperty.call(rule, "required")) {
+                requiredValidation = rule;
+            } else {
+                fieldValidation.push(rule);
+            }
+        });
+        const newInternalId = SimpleIdGenerator.generate();
+
+        const translatedElements = {
```
```suggestion
        const parsedElements = {
```

weili-govtech (changes requested):

weili-govtech:
observed another issue where the conditional rendering lists itself as the source field
1. add an email field
2. add a short text field
3. add a conditional rendering rule to email field
4. save it
5. switch to preview form
6. switch back
7. click email field
8. open up conditional rendering rules
9. click on field dropdown

image

src/components/element-editor/conditional-rendering/conditional-rendering.tsx
```diff
@@ -149,7 +149,7 @@ export const ConditionalRendering = () => {
             title="Conditional Rendering"
             buttonLabel="condition"
             disabledButton={
-                getElementOptions().length === 0 || invalidAndEmptyFields
+                Object.values(elements).length === 1 || invalidAndEmptyFields
```
weili-govtech: this prevents the button from being disabled even if there are invalid fields.

shouldnt it be disabled if there are no other elements?

src/translator/text-based-field.ts
weili-govtech, line 1: rename `translatedElements` as `parsedElements`

weili-govtech (changes requested):

weili-govtech:

weili-govtech (approved):

PR #80, opened 2024-07-04
[CCUBE-1431][SR] Enabled the reordering of the lements to be shown in the form perview mode.

ShIrInRuTh:
**Changes**
- Enabled the reordering of the lements to be shown in the form perview mode.

**Additional information**

-   You may refer to this [CCUBE-1431](https://jira.ship.gov.sg/browse/CCUBE-1431)
-   Bug fix for CCUBE-1431 based on deskcheck

### DIFF
--- a/src/context-providers/builder/hook.ts
+++ b/src/context-providers/builder/hook.ts
@@ -20,7 +20,7 @@ export const useBuilder = () => {
                 payload: orderedIdentifiers,
             });
         },
-        [state.orderedIdentifiers]
+        []
     );
 
     const addElement = useCallback(
--- a/src/context-providers/builder/hook.ts
+++ b/src/context-providers/builder/hook.ts
@@ -15,21 +15,12 @@ export const useBuilder = () => {
 
     const updateOrderedIdentifiers = useCallback(
         (orderedIdentifiers: IElementIdentifier[]) => {
-            const newElements = {};
-            orderedIdentifiers.map((value) => {
-                newElements[value.internalId] =
-                    state.elements[value.internalId];
-            });
-
             dispatch({
                 type: "update-ordered-identifiers",
-                payload: {
-                    orderedIdentifiers,
-                    elements: newElements,
-                },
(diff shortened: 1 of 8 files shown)


src/context-providers/builder/hook.ts
```diff
@@ -15,12 +15,21 @@ export const useBuilder = () => {
 
     const updateOrderedIdentifiers = useCallback(
         (orderedIdentifiers: IElementIdentifier[]) => {
+            const newElements = {};
+            orderedIdentifiers.map((value) => {
+                newElements[value.internalId] =
+                    state.elements[value.internalId];
+            });
```
weili-govtech: instead of reordering the elements here, a more complete solution will be to do so in `generateSchema()`

this is because there are other actions updating the ordered identifiers.

though irrelevant now, `.map()` is not appropriate here because we expect a 1-to-1 transformation with it. `.reduce()` would have been more suitable

weili-govtech (changes requested):

src/context-providers/builder/hook.ts
```diff
@@ -20,7 +20,7 @@ export const useBuilder = () => {
                 payload: orderedIdentifiers,
             });
         },
-        []
+        [state.orderedIdentifiers]
```
weili-govtech: why is this needed?

stories/translator/doc-elements.tsx
```diff
@@ -54,10 +58,18 @@ export const DocElement = () => {
...
+        }
     };
 
     const handleSchemaViewButton = () => {
         setPageMode("schema-mode");
+        if (formBuilderRef.current) {
+            const generatedSchema = formBuilderRef.current?.generateSchema();
+            setSchema(generatedSchema);
+        }
     };
```
weili-govtech: I dont quite understand why the need to make these changes, what are we fixing with these changes?

weili-govtech (changes requested):

weili-govtech (approved):

PR #84, opened 2024-07-11
CCUBE-1480, CCUBE-1466

adenteo:
This PR contains commits for both CCUBE-1480 and CCUBE-1466

**Changes**

**CCUBE-1480**
- Refactor validation, conditional rendering and prefill to use `useFieldArray`
  - Remove need for `useEffect`/`useState`
  - Replace custom remove/add logic with built in `remove` and `append`
  - Remove shouldUnregister since not recommended for useFieldArray
- Shift cross-cutting logic of form to element editor
- Update element to clean state when unfocused
- Add default values to `useForm`
- Edit form reset logic upon submission
- Add `shouldUnregister` to all fields
- Add `setTimeout` to `reset()`

**CCUBE-1466**
- Show correct validation types based on element type
- Mark form as dirty and clear validation once element type is touched
- Add default values to watch to fix failing test cases

**Additional information**

-   You may refer to [CCUBE-1466](https://jira.ship.gov.sg/browse/CCUBE-1466) and [CCUBE-1480](https://jira.ship.gov.sg/browse/CCUBE-1480)

### DIFF
--- a/src/components/element-editor/element-editor.tsx
+++ b/src/components/element-editor/element-editor.tsx
@@ -8,32 +8,40 @@ import {
 } from "./element-editor.styles";
 import { Prefill } from "./prefill";
 import { Validation } from "./validation/validation";
+import { useEffect } from "react";
+import { useFormContext } from "react-hook-form";
+import { IBaseTextBasedFieldValues } from "src/schemas";
 
 export const ElementEditor = () => {
     // =========================================================================
     // CONST, STATE, REF
     // =========================================================================
-    const { focusedElement } = useBuilder();
+    const { focusedElement, updateFocusedElement } = useBuilder();
 
-    // =============================================================================
-    // RENDER FUNCTIONS
-    // =============================================================================
+    const {
+        formState: { isDirty },
+        getFieldState,
+        getValues,
+    } = useFormContext<IBaseTextBasedFieldValues>();
 
-    const renderAlert = () => {
-        if (!focusedElement.isDirty) {
-            return <></>;
-        } else {
-            return (
-                <SaveChangesAlert type="warning" showIcon>
-                    To reflect changes on preview, save changes first.
-                </SaveChangesAlert>
-            );
+    const { isTouched: isTypeTouched } = getFieldState("type");
+
+    useEffect(() => {
+        // Once the element type is touched, that element should be considered dirty until it is saved or its changes are discarded.
+        if (isTypeTouched) {
+            updateFocusedElement(true);
+            return;
         }
-    };
+        updateFocusedElement(!!isDirty);
+    }, [isDirty, updateFocusedElement]);
 
     return (
         <Wrapper>
-            {renderAlert()}
+            {focusedElement.isDirty && (
+                <SaveChangesAlert type="warning" showIcon>
+                    To reflect changes on preview, save changes first.
+                </SaveChangesAlert>
+            )}
             <AccordionWrapper>
                 <BasicDetails />
             </AccordionWrapper>
--- a/src/components/element-editor/element-editor.tsx
+++ b/src/components/element-editor/element-editor.tsx
@@ -1,4 +1,7 @@
+import { useEffect } from "react";
+import { useFormContext } from "react-hook-form";
 import { useBuilder } from "src/context-providers";
+import { IBaseTextBasedFieldValues } from "src/schemas";
 import { BasicDetails } from "./basic-details";
 import { ConditionalRendering } from "./conditional-rendering";
 import {
@@ -8,9 +11,6 @@ import {
 } from "./element-editor.styles";
 import { Prefill } from "./prefill";
 import { Validation } from "./validation/validation";
-import { useEffect } from "react";
-import { useFormContext } from "react-hook-form";
-import { IBaseTextBasedFieldValues } from "src/schemas";
 
 export const ElementEditor = () => {
     // =========================================================================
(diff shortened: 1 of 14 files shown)


src/components/element-editor/element-editor.tsx
```diff
@@ -8,32 +8,39 @@ import {
...
-        if (!focusedElement.isDirty) {
-            return <></>;
-        } else {
-            return (
-                <SaveChangesAlert type="warning" showIcon>
-                    To reflect changes on preview, save changes first.
-                </SaveChangesAlert>
-            );
+    const { isTouched: isTypeTouched } = getFieldState("type");
+
```
weili-govtech, line 27: add comment divider for use effects

```
// EFFECTS
```

src/components/element-editor/element-editor.tsx
```diff
@@ -8,32 +8,39 @@ import {
 } from "./element-editor.styles";
 import { Prefill } from "./prefill";
 import { Validation } from "./validation/validation";
+import { useEffect } from "react";
+import { useFormContext } from "react-hook-form";
+import { IBaseTextBasedFieldValues } from "src/schemas";
```
weili-govtech: imports should be organised in alphabetical order, you can do so via the vscode command palette

src/components/element-editor/element-editor.tsx
```diff
@@ -8,32 +8,39 @@ import {
...
+
+    useEffect(() => {
+        // Once the element type is touched, that element should be considered dirty until it is saved or its changes are discarded.
+        if (isTypeTouched) {
+            updateFocusedElement(true);
+            return;
         }
-    };
+        updateFocusedElement(!!isDirty);
+    }, [isDirty, updateFocusedElement]);
```
weili-govtech, line 40: add comment divider for renders

```
// RENDER FUNCTIONS
```

src/context-providers/builder/provider.tsx
```diff
@@ -60,6 +60,7 @@ export const builderReducer = (
             break;
         }
         case "remove-focused-element": {
+            state.focusedElement.isDirty = false;
```
weili-govtech: why do we need to change this before changing focus?

src/components/side-panel/side-panel.tsx
```diff
@@ -30,10 +30,22 @@ export const SidePanel = ({ offset, onSubmit }: IProps) => {
...
     const methods = useForm({
         mode: "onTouched",
+        defaultValues: {
+            type: "",
+            label: "",
+            required: true,
+            requiredErrorMsg: "",
+            id: "",
+            placeholder: "",
+        },
```
weili-govtech: can we set the default values in each `Controller` instead? because the fields rendered differ between element types

weili-govtech (changes requested):

src/context-providers/builder/provider.tsx
adenteo: I realised there is no need for this since we are already setting `focusedElement` to `null`. However I did notice that if we dirty a form and then discard changes and click on a new element, on first render of `ElementEditor`, form would still be dirty since we did not reset it when discarding changes. It is set to `false` in the `useEffect` hook on 2nd render in `SidePanel`, but there will be a quick flicker in the side panel from dirty to clean. To prevent that, I can add a `reset` when we discard changes. Will have to edit the test cases too. Should I go ahead with this or is the flicker negligible?

src/components/side-panel/side-panel.tsx
adenteo: From RHF [docs](https://react-hook-form.com/docs/usecontroller), in order to use defaultValue in `Controller`, we need to define `defaultValue` in `useForm`. 

We can remove the default value for the rest of the fields actually, but for `requiredErrorMsg`, if we don't, when it remounts, `defaultValue` for `requiredErrorMsg` is set to `undefined` in the form. Having `defaultValue` set in the `Controller` will only set the value of the field, but the form will still expect `undefined` in order to be clean.

As for the rest of the fields, I feel like we also might not need to set a `defaultValue` in each of them. This is because whenever the form is shown, we do a `reset` with the new default values, which would overwrite any `defaultValue` set within the `Controller` if I am not wrong.

src/context-providers/builder/provider.tsx
weili-govtech: I didnt observe this behaviour though, the `isDirty` state is only present in the `focusedElement` so on switching the focused element, wont that state be lost hence become falsy?

src/components/side-panel/side-panel.tsx
weili-govtech: ok fair point, let's go with this then

src/context-providers/builder/provider.tsx
adenteo: Yes, the dirty state of the `focusedElement` does become falsy. However:

In line 38 of `ElementEditor`,

`updateFocusedElement(!!isDirty);`

updates the focused element's dirty state **based on the form's dirty state**. When `ElementEditor` is rendered again on re-focus after discarding changes, the form is actually still dirty, since we don't reset it when discarding changes. If we `console.log(isDirty)` here, we will realise on first render it's setting the dirty state of `focusedElement` to `true`, then `false` on second render.

weili-govtech: I see, thanks for spotting this, yes using `reset` is more appropriate here, trouble you to add this of any discarding of changes

src/components/element-editor/validation/validation.tsx
```diff
@@ -55,14 +58,13 @@ export const Validation = () => {
     };
 
     const getPopoverMessage = () => {
-        const invalidAndEmptyFields = checkIsValid();
-        if (invalidAndEmptyFields) {
+        if (hasInvalidAndEmptyfields()) {
```
weili-govtech:
```suggestion
        if (hasInvalidAndEmptyFields()) {
```

weili-govtech (changes requested):

weili-govtech:
- need to resolve failed tests
- need to resolve merge conflicts

weili-govtech (approved):

PR #89, opened 2024-07-16
CCUBE-1491

ShIrInRuTh:
**Changes**

- Added new fields for the long text field
- Added new typings for the long text field
- Added validation for the long text field
- Amended the element object generator to add the new fields for the long text field
- Added unit tests to test the long text field
- Amended failing unit tests.
- Added text for checking if the alert will be shown for long text validation.

**Additional information**

-   You may refer to this [CCUBE-1491](https://jira.ship.gov.sg/browse/CCUBE-1491)
-   This is branched out of CCUBE-1489

### DIFF
--- a/src/components/element-editor/basic-details/basic-details.tsx
+++ b/src/components/element-editor/basic-details/basic-details.tsx
@@ -79,7 +79,7 @@ export const BasicDetails = () => {
                                 label="Element Name"
                                 rows={1}
                                 placeholder="Element Name"
-                                value={field.value}
+                                value={field.value as string}
                                 onChange={(e) => {
                                     field.onChange(e.target.value);
                                 }}
--- a/src/components/element-editor/basic-details/basic-details.tsx
+++ b/src/components/element-editor/basic-details/basic-details.tsx
@@ -19,12 +19,11 @@ export const BasicDetails = () => {
     // =========================================================================
     // CONST, STATE, REFS
     // =========================================================================
-    const { focusedElement, updateFocusedElement, elements } = useBuilder();
+    const { focusedElement } = useBuilder();
     const {
         control,
         formState: { errors },
         watch,
-        reset,
         setValue,
     } = useFormContext<IBaseTextBasedFieldValues | ITextareaFieldAttributes>();
     const element = focusedElement.element;
--- a/src/components/element-editor/basic-details/basic-details.tsx
+++ b/src/components/element-editor/basic-details/basic-details.tsx
@@ -4,7 +4,11 @@ import { useEffect } from "react";
 import { Controller, useFormContext } from "react-hook-form";
 import { IconDropdown } from "src/components/common/icon-dropdown";
 import { TogglePair } from "src/components/common/toggle-pair/toggle-pair";
-import { EElementType, useBuilder } from "src/context-providers";
+import {
+    EElementType,
+    ITextareaFieldAttributes,
+    useBuilder,
+} from "src/context-providers";
 import { IBaseTextBasedFieldValues } from "src/schemas";
 import {
     FieldEditorAccordionItem,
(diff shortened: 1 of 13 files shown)


src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -48,6 +52,8 @@ export const BasicDetails = () => {
         });
     }, []);
 
+    console.log("element:", element, elements);
```
weili-govtech: remove errant console.log

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -104,7 +110,10 @@ export const BasicDetails = () => {
                         control={control}
                         render={({ field }) => (
                             <TogglePair
-                                label="Mandatory field"
+                                label={{
+                                    mainLabel: "Mandatory field",
+                                    subLabel: "",
```
weili-govtech: remove `subLabel` since it is not used

src/schemas/text-based-fields.ts
```diff
@@ -95,13 +92,25 @@ const generateValidationSchema = (elementType: EElementType) => {
...
-                            is:
-                                ELEMENT_VALIDATION_TYPES["Text field"][
-                                    EElementType.TEXT
-                                ].validationTypes[1] ||
-                                ELEMENT_VALIDATION_TYPES["Text field"][
-                                    EElementType.TEXT
-                                ].validationTypes[2],
+                            is: ELEMENT_VALIDATION_TYPES["Text field"][
+                                EElementType.TEXT
+                            ].validationTypes[1],
```
weili-govtech: instead of checking by array position, can we change to find the presence of the validation rule in the array?

make the same change for line 47-49, 75-77, 111-113, 184-185

weili-govtech (changes requested):

src/schemas/text-based-fields.ts
```diff
@@ -75,9 +74,15 @@ const generateValidationSchema = (elementType: EElementType) => {
...
-                                EElementType.TEXT
-                            ].validationTypes[0],
+                            is: (value: string) => {
+                                const validationTypes =
+                                    ELEMENT_VALIDATION_TYPES["Text field"][
+                                        EElementType.TEXT
+                                    ].validationTypes;
+                                return validationTypes
+                                    .slice(0, 1)
+                                    .includes(value);
```
weili-govtech: this makes no difference right? you are still checking by array position, the risk here is if the array order is changed, all the logic here will break.

you should check for the presence of the validation rule from the array, e.g. check for `Custom regex` rule here

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -16,13 +20,13 @@ export const BasicDetails = () => {
     // =========================================================================
     // CONST, STATE, REFS
     // =========================================================================
-    const { focusedElement, updateFocusedElement } = useBuilder();
+    const { focusedElement, updateFocusedElement, elements } = useBuilder();
```
weili-govtech: `updateFocusedElement` and `elements` are not used

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -15,13 +19,14 @@ export const BasicDetails = () => {
     // =========================================================================
     // CONST, STATE, REFS
     // =========================================================================
-    const { focusedElement } = useBuilder();
+    const { focusedElement, updateFocusedElement, elements } = useBuilder();
     const {
         control,
         formState: { errors },
         watch,
+        reset,
```
weili-govtech: `reset` is not used

weili-govtech (changes requested):

weili-govtech:
do fix the failing tests

src/schemas/text-based-fields.ts
```diff
@@ -95,13 +101,16 @@ const generateValidationSchema = (elementType: EElementType) => {
...
-                                ].validationTypes[2],
+                            is: (value: string) => {
+                                const validationTypes =
+                                    ELEMENT_VALIDATION_TYPES["Text field"][
+                                        EElementType.TEXT
+                                    ].validationTypes;
+                                return (
+                                    validationTypes.includes(value) &&
+                                    value !== "Custom regex"
+                                );
```
weili-govtech: apologies, on second look I think it's better to handle each validation type explicitly

also, remove the `otherwise` handler because it will never run

```suggestion
                                return (
                                    validationTypes.includes(value) &&
                                    (value === "Minimum length" ||
                                        value === "Maximum length")
                                );
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #96, opened 2024-07-23
[CCUBE-1436][SR] Amended the bug causing issues in toggling back to form builder mode

ShIrInRuTh:
**Changes**

-  Amended the generating for schema, for mandatory field.
- Amended label type error, causing build failure.

**Additional information**

-   You may refer to this [CCUBE-1436](https://jira.ship.gov.sg/browse/CCUBE-1436)
-   This is a bugfix for a bug that @adenteo pointed out the inability to return to form builder mode ans validation throws an bug.

### DIFF
--- a/src/context-providers/builder/element.types.ts
+++ b/src/context-providers/builder/element.types.ts
@@ -1,5 +1,3 @@
-import { IComplexLabel } from "@lifesg/web-frontend-engine/components/fields";
-
 /**
  * Refer to the uiType in the Frontend Engine when setting
  * the value.
--- a/src/context-providers/builder/element.types.ts
+++ b/src/context-providers/builder/element.types.ts
@@ -62,7 +62,7 @@ export interface IColumns {
 }
 
 export interface IBaseFieldAttributes extends IBaseAttributes {
-    label: string | IComplexLabel;
+    label: string;
     required: boolean;
     requiredErrorMsg?: string | undefined;
     columns: IColumns;
--- a/src/translator/text-based-field.ts
+++ b/src/translator/text-based-field.ts
@@ -165,7 +165,7 @@ export namespace TextBasedField {
     ) => {
         const { showIf, uiType, validation, label, ...rest } = element;
 
-        let requiredValidation: IYupValidationRule = { required: false };
+        let requiredValidation: IYupValidationRule = {};
         const fieldValidation = [];
 
         validation?.forEach((rule) => {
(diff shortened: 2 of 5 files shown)


src/context-providers/builder/element.types.ts
```diff
@@ -62,7 +62,7 @@ export interface IColumns {
 }
 
 export interface IBaseFieldAttributes extends IBaseAttributes {
-    label: string | IComplexLabel;
+    label: string;
```
weili-govtech, line 63: remove the import of `IComplexLabel`

src/translator/text-based-field.ts
```diff
@@ -76,9 +76,9 @@ export namespace TextBasedField {
     const createValidationObject = (element: TElement) => {
         const validation: IYupValidationRule[] = [];
 
-        if (element.required) {
+        if (element.required !== undefined) {
             validation.push({
-                required: true,
+                required: element.required,
```
weili-govtech: we should assign the validation only if the field is required, is this causing an issue?

src/translator/text-based-field.ts
weili-govtech, line 1: remove unused import of `TRenderRules`

tests/util/translator/translator.spec.tsx
```diff
@@ -41,6 +41,11 @@ describe("Translator", () => {
                             mainLabel: "Email address",
                         },
                         uiType: EElementType.EMAIL,
+                        validation: [
+                            {
+                                required: false,
+                            },
+                        ],
```
weili-govtech: related to previous point, this isnt necessary

weili-govtech (changes requested):

src/translator/text-based-field.ts
weili-govtech: I think you missed this comment?

src/translator/text-based-field.ts
```diff
@@ -166,10 +166,10 @@ export namespace TextBasedField {
     ) => {
         const { showIf, uiType, validation, label, ...rest } = element;
 
-        let requiredValidation: IYupValidationRule = {};
+        let requiredValidation: IYupValidationRule = { required: false };
```
weili-govtech: this isnt necessary, you can cast the value as boolean on line 186

```
required: !!requiredValidation.required,
```

with this, even if `requiredValidation` is assigned an invalid value subsequently, this will still convert it in true / false

weili-govtech (approved):

PR #102, opened 2024-07-31
CCUBE-1490 Schema Parsing

adenteo:
This branch focuses on implementing the parsing of schema to element for text-based fields.

**Changes**
- Add enum for FEE validation type
- Add schema parsing functionality for text-based elements
- Edit validation types to refer to exact type 
- Fix failing test cases.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1490)

### DIFF
--- a/src/translator/text-based-field.ts
+++ b/src/translator/text-based-field.ts
@@ -8,11 +8,11 @@ import {
 import {
     EElementType,
     EValidationType,
+    EValidationTypeFEE,
     IValidation,
     TElement,
     TTextBasedElement,
 } from "src/context-providers";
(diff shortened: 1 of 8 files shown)


src/translator/text-based-field.ts
```diff
@@ -123,6 +120,48 @@ export namespace TextBasedField {
         return validationObj;
     };
 
+    export const parseTextBasedFieldValidation = (
```
weili-govtech, line 123: ideally we want to:
- differentiate between element types 
- aside from checking for the presence of the key, check if it is true 
but I will let these changes get through first since they will be refactored

weili-govtech (approved):

PR #106, opened 2024-08-06
CCUBE-1510-2

adenteo:
This PR focuses on refactoring the basic details into individual components for each element type.  It also ensures the correct behaviour when toggling between element types.

**Changes**
- Create separate basic details component for each type.
- Ensure current values are properly set when changing types.
- Refactor basic details and pills test cases
- Remove shouldUnregister for common fields. 
- Clear preselectedValue when toggling from dropdown. 
- Change empty value of preselectedValue of dropdown.
- Fix failing test cases
- Rename textarea to long-text
- Edit test cases and remove unnecessary code.
- Reset form values when changing element types.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1510)


src/components/element-editor/basic-details/basic-details.styles.tsx
weili-govtech, line 1: we try to keep only the styles used by `BasicDetails` in this file

- remove unused `MandatoryFieldBox`
- moved `StyledTextarea` to a new `common/common.styles.ts`, we will share the styled-components in the same file for now until it gets too big
- move `ToggleWrapper` into `TogglePair`, move the styles into `toggle-pair.styles.tsx`

other than that, can rename `FieldEditorAccordionItem` as `AccordionItem` to follow the `AccordionWrapper` naming format

src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -1,90 +1,58 @@
...
-    TypeField,
-} from "./common";
-import { DropdownItems } from "./dropdown-items/dropdown-items";
-import { DropdownPreselectedValue } from "./dropdown-preselected-value";
+import { ContactBasicDetails } from "./contact";
+import { DropdownBasicDetails } from "./dropdown";
+import { EmailBasicDetails } from "./email";
+import { NumericBasicDetails } from "./numeric";
+import { TextBasicDetails } from "./text";
+import { LongTextBasicDetails } from "./long-text";
```
weili-govtech: move them into `elements/` folder and do barrel export

src/components/element-editor/basic-details/common/type-field.tsx
```diff
@@ -27,8 +27,18 @@ export const TypeField = () => {
...
+        }
         selectElementType(value);
         setValue("validation", []);
+        /** The purpose of this setValue is to deliberately clear the preselectedValue
+         * when the input type of the preselectedValue is different. Currently,
+         * the preselectedValue is different only when the input type is dropdown.
+         */
+        if (field.value === EElementType.DROPDOWN) {
+            setValue("preselectedValue", "");
+        }
```
weili-govtech: is it possible to move this logic into `DropdownBasicDetails`?
because this logic has cross cutting concerns and it may not be obvious enough to be kept here

tests/components/common/pills/pills.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+            disconnect: jest.fn(),
+        }));
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+        jest.resetAllMocks();
+    });
+
+    it("should render the pill items and list position field if pills field is toggled from false to true", async () => {
```
weili-govtech: split into multiple tests to narrow the scope of each test
- test that pills are false by default and no pill items are rendered
- test for conditional rendering of pill items

tests/components/element-editor/basic-details/basic-details.spec.tsx
weili-govtech, line 1: now that `BasicDetails` don't render the individual fields, the tests for the conditional rendering of the fields are not needed anymore. instead it should be testing the conditional rendering of the element-specific basic details component. you may need to use testid to do this.

weili-govtech (changes requested):

src/components/element-editor/basic-details/common/type-field.tsx
adenteo: ~This logic currently runs when `type` is being toggled from `EElementType.DROPDOWN` to another element. I did try running it on unmounting of `DropdownPreselectedValue` with a `useEffect`, but faced the same issue of the value being retained in the input field. It can't be shifted into `TextBasedPreselectedValue` too since there's no way of knowing if dropdown type was selected previously~

~Karling did mention that we should clear all fields when type is changed, so this logic will probably be edited too in the future~

EDIT: Field values are now all reset upon changing type

weili-govtech: will it work if we have a useEffect with type in dependencies in `BasicDetails`? the useEffect will also check if the form is dirty, if it is, it will perform the reset

tests/components/element-editor/dropdown-items/dropdown-items.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import { fireEvent, render, screen } from "@testing-library/react";
 import "jest-canvas-mock";
 import { act } from "react-dom/test-utils";
-import { DropdownItems } from "src/components/element-editor/basic-details/dropdown/dropdown-items";
+import { DropdownItems } from "src/components/element-editor/basic-details/elements/dropdown";
```
weili-govtech:
```suggestion
import { DropdownItems } from "src/components/element-editor/basic-details/elements";
```

tests/components/element-editor/basic-details/basic-details.spec.tsx
```diff
@@ -18,75 +18,139 @@ describe("BasicDetails", () => {
...
     });
-
-    describe("rendering label & required error message fields", () => {
-        it("should render label if element has label property", async () => {
-            renderComponent({
-                builderContext: {
-                    focusedElement: MOCK_FOCUSED_ELEMENT,
-                    selectedElementType: EElementType.EMAIL,
-                },
+    describe.each`
```
weili-govtech: since it contains 1 test only, use `it.each` instead

tests/components/element-editor/basic-details/basic-details.spec.tsx
```diff
@@ -18,75 +18,139 @@ describe("BasicDetails", () => {
...
+
+            let elementName = screen.getByTestId("label-field");
+            let errorMessageField = screen.getByTestId(
+                "required-error-message-field"
+            );
+            let elementId = screen.getByTestId("id-field");
+
+            expect(elementName).toHaveValue("Email address");
+            expect(errorMessageField).toHaveValue("This is a required field.");
+            expect(elementId).toHaveValue("email-field");
```
weili-govtech: you can move these assertions into a separate test to make sure the initial state is assigned properly

in the subsequent tests, you can skip these assertions since they have been covered

weili-govtech (changes requested):

src/components/element-editor/basic-details/common/type-field.tsx
adenteo: yes a useEffect in `BasicDetails` would work. Will shift it there. As for the 2nd part, can I clarify what you mean? Currently based on Karling's mockup, we will reset the fields even if the form is clean and in saved state.

weili-govtech: because I am proposing to create a useEffect with only type as dependency:
```
 useEffect(() => {
  // reset
}, [type]);
```

we cant reset the form based on type alone because it will also run every time I select or switch an element in the main panel. but since the form will be dirtied on changing the type, we can reset the form based on that.

tests/components/element-editor/basic-details/basic-details.spec.tsx
```diff
@@ -19,232 +19,155 @@ describe("BasicDetails", () => {
...
 
-            const submitButton = screen.getByText("Submit");
-            fireEvent.click(submitButton);
+            // Check that the fields are cleared
+            waitFor(() => {
+                descriptionField = screen.getByTestId("description-field");
+                placeholderField = screen.getByTestId("placeholder-field");
+                preselectedValueField = screen.getByTestId(
+                    "preselected-value-field"
+                );
```
weili-govtech: not needed, can rely on the initial queries at the beginning of this test

src/components/element-editor/basic-details/common/type-field.tsx
adenteo: I see. I missed that out. Thank you for the clarification!

tests/components/element-editor/basic-details/basic-details.spec.tsx
adenteo: By using the initial queries, I don't seem to be getting the updated values after the `useEffect` runs.

PR #107, opened 2024-08-06
CCUBE-1511-1

adenteo:
This branch is solely focusing on creating new common components for validation, along with their test cases. These components are not being used yet.

In the next iteration, they will be implemented, and validation will be further broken down into different custom components for element types that have different validation, such as email address and numeric fields.

**Changes**
- Create common validation components
- Add test cases for new validation components

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1511)


src/components/element-editor/validation/common/validation-child.tsx
```diff
@@ -0,0 +1,38 @@
+import { ChildEntry } from "src/components/common";
+import {
+    ValidationErrorMessage,
+    ValidationRule,
+    ValidationType,
+} from "../common/";
+import { FieldWrapper } from "../validation.styles";
```
weili-govtech: move over to `validation/common/` folder and rename as `validation.child.styles.ts`

src/components/element-editor/validation/common/validation-type.tsx
```diff
@@ -0,0 +1,46 @@
...
+                return (
+                    <Form.Select
+                        {...fieldWithoutRef}
+                        placeholder="Select"
+                        selectedOption={fieldWithoutRef.value}
+                        options={options}
+                        disabled={disabled}
+                        onSelectOption={(option) => {
+                            field.onChange(option);
+                        }}
```
weili-govtech:
```suggestion
                        onSelectOption={field.onChange}
```

src/components/element-editor/validation/common/validation-child.tsx
```diff
@@ -0,0 +1,38 @@
+import { ChildEntry } from "src/components/common";
+import {
+    ValidationErrorMessage,
+    ValidationRule,
+    ValidationType,
+} from "../common/";
```
weili-govtech:
```suggestion
} from ".";
```

src/components/element-editor/validation/common/validation-error-message.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const ValidationErrorMessage = ({ index }: IProps) => {
+    // ===========================================================================
+    // CONST, STATE, REF
+    // ===========================================================================
+    const {
+        formState: { errors },
+        control,
+    } = useFormContext<TOverallTextBasedSchema>();
```
weili-govtech: instead of `TOverallTextBasedSchema`, we can have a new type to cover all schemas with validation. this way if new elements are added, devs wont miss adding the support to these components

src/components/element-editor/validation/common/validation-child.tsx
weili-govtech, line 1: may I check again how is the component meant to be used? is it meant to cover the standard validations that come with a rule selection dropdown, rule input and error message input? for the customised ones, they will be in their element-specific components?

weili-govtech (changes requested):

src/components/element-editor/validation/common/validation-child.tsx
adenteo: Yes. This is meant to cover the standard ones. Elements that require customised `validation-child`s would be created in their folders.

src/components/element-editor/validation/common/validation-child.tsx
weili-govtech: something to note, the "x" in `tsx` format represents `xml`, we use it when tags are used, e.g. `<FieldWrapper />`, so most styles files can drop the "x" since they wont contain any tags.

adenteo: Got it. Will rename the validation style file in this PR. For the rest of the style files outside of this PR, perhaps I can create another PR in the future to remove all unnecessary "x"s.

PR #108, opened 2024-08-12
CCUBE-1511-2

adenteo:
**Changes**
- Add custom validation child and validation rule.
- Add validation test cases for new custom components
- Fix error messages in test cases
- Shift custom validation components into common elements folder. 
- Modify test cases.
- Create custom validation child for text and contact. 
- Edit validation child logic.
- Organise imports

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1511)


src/components/element-editor/basic-details/basic-details.tsx
```diff
@@ -69,6 +70,7 @@ export const BasicDetails = () => {
                                     shouldDirty: true,
                                 });
                                 setValue("validation", []);
+                                clearErrors("validation");
```
weili-govtech: replace with `resetField`

side note: remember to move this logic to a useEffect as proposed in https://github.com/LifeSG/web-form-builder/pull/106#discussion_r1708421548

src/components/element-editor/validation/elements/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./email";
+export * from "./text";
+export * from "./long-text";
+export * from "./numeric";
```
weili-govtech:
```suggestion
export * from "./contact";
export * from "./email";
export * from "./long-text";
export * from "./numeric";
export * from "./text";

```

src/components/element-editor/validation/validation.tsx
```diff
@@ -124,17 +119,67 @@ export const Validation = () => {
...
+                            onDelete={() => handleDelete(index)}
+                        />
+                    );
+                case EElementType.NUMERIC:
+                    return (
+                        <NumericValidationChild
+                            key={field.id}
+                            index={index}
+                            options={options}
+                            disabled={options.length === 1}
```
weili-govtech: as admin can add up to 3 rules, this is causing all the rule selection to be disabled from 2nd rule onwards. since admin cant add beyond 3 rules and the added options are removed, we can remove this logic.

weili-govtech (changes requested):

src/components/element-editor/basic-details/basic-details.tsx
adenteo: I've changed it to use `resetField`. However, when we use `resetField`, we would have to set a `defaultValue` of `[]` to clear the validation, since `validation` will be reset with it's `defaultValues` if it was saved previously.

This will also reset the actual `defaultValue` of `validation` to `[]` and mark it as clean.

In this case it should be fine, since the `type` is dirtied so the form will still stay dirtied.

weili-govtech (approved):

PR #111, opened 2024-08-13
CCUBE-1512-1

adenteo:
**Changes**
- Create common functions and helpers for generate schema logic.
- Create separate schema generators for each element type.
- Add test cases for each element schema generator.
- Edit test cases for translator.
- Extract validationType yup schema.
- Remove unused element types.
-  Rename functions and file in generator to start with generate.
- Add base schema test cases.

**Additional information**
-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1512)


src/schemas/text-based-fields.ts
```diff
@@ -46,14 +46,26 @@ yup.addMethod(yup.string, "isNumber", function (message) {
     });
 });
 
+const validationTypeSchema = () =>
+    yup
+        .string()
+        // Convert empty strings to undefined to trigger the required validation
+        .transform((value, originalValue) => {
+            return originalValue === "" ? undefined : value;
+        })
```
weili-govtech, line 55: is it necessary to convert to undefined? required validation works for empty string too

src/translator/generate/index.ts
weili-govtech, line 1: do barrel export for `elements/` only because only they are used outside the folder

when importing within the same folder, avoid import via barrel exports because it may cause circular imports

weili-govtech (changes requested):

src/schemas/text-based-fields.ts
adenteo: It seems to be necessary. Without it, the `oneOf` validation would trigger and return an error message.

src/translator/generate/elements/numeric/numeric-schema-generator.ts
```diff
@@ -0,0 +1,62 @@
+import {
+    EValidationType,
+    INumericField,
+    IValidation,
+} from "src/context-providers";
+import { ISchemaValidation } from "../..";
```
weili-govtech:
```suggestion
import { ISchemaValidation } from "../../types";
```

src/schemas/text-based-fields.ts
weili-govtech: looks like `oneOf` takes higher precedence over `required`, there are other workarounds but I think this is sufficient too so I am resolving this issue

weili-govtech (changes requested):

weili-govtech:
need to fix the import in `validation-generators.ts` too

weili-govtech (approved):

PR #116, opened 2024-08-20
CCUBE-1512-2

adenteo:
This branch extracts the `placeholder` attribute from the base schema generator since realising not all element types have placeholder (Yes/No element as of now). It also covers the renaming of consts in the test cases related to schema generating from UPPER_SNAKE_CASE  to camelCase.

**Changes**

- Extract placeholder from base schema generator. 
- Add test cases for placeholder.
- Remove default case from setDefaultValidationType.
- Rename consts used in functions to be camelCase.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1512)

### DIFF
--- a/tests/util/translator/generate/elements/dropdown/dropdown-schema-generator.spec.tsx
+++ b/tests/util/translator/generate/elements/dropdown/dropdown-schema-generator.spec.tsx
@@ -20,8 +20,8 @@ describe("DropdownSchemaGenerator", () => {
         const placeholder = "This is a placeholder";
 
         const mockElement: IDropdown = {
-            label,
-            id: elementId,
+            label: LABEL,
+            id: ELEMENT_ID,
             internalId: "dropdown-field",
             type: EElementType.DROPDOWN,
             required: false,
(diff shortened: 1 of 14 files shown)


tests/util/translator/generate/elements/dropdown/dropdown-schema-generator.spec.tsx
```diff
@@ -63,3 +92,9 @@ describe("DropdownSchemaGenerator", () => {
         expect(generatedSchema).toStrictEqual(expectedSchema);
     });
 });
+
+// =============================================================================
+// HELPERS
+// =============================================================================
+const elementId = "mockId123";
+const label = "dropdown-field";
```
weili-govtech: as these constants are outside any functions, we want to use UPPER_SNAKE_CASE 

trouble you to fix for the other tests as well

```suggestion
const ELEMENT_ID = "mockId123";
const LABEL = "dropdown-field";
```

weili-govtech (changes requested):

tests/util/translator/generate/elements/dropdown/dropdown-schema-generator.spec.tsx
adenteo: Fixed for generators. Will do so for parsers in corresponding branch.

weili-govtech (approved):

PR #121, opened 2024-08-22
CCUBE-1523

adenteo:
This branch introduces the `reset-fields-modal` that will be shown when users change element type within the element editor panel.

**Changes**
- Add ResetFieldsModal and corresponding types
- Edit basic details test cases
- Fix circular depencies in modal
- Add test cases for reset-fields-modal

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1523)


src/components/common/modal/reset-fields-modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    return (
+        <GenericModal type={modal?.type} onClose={onClose}>
+            <ErrorDisplay
+                type="confirmation"
+                description={
+                    "You've changed the element type. This will reset all fields to match the new element type."
+                }
+                title={
+                    <ModalDisplayTitle weight={"semibold"}>
```
weili-govtech, line 45: move the styled-component to `modals.styles` since a few components are using it

src/components/element-editor/basic-details/common/type-field.tsx
```diff
@@ -29,10 +40,29 @@ export const TypeField = () => {
...
+
+        showModal(resetFieldsModal);
     };
 
+    // This useEffect is used to trigger a re-render of the IconDropdown component when the user decides not to reset the fields, such that the dropdown is updated with the correct type
+    useEffect(() => {
+        if (localType === null) {
+            setLocalType(selectedElementType);
+        }
+    }, [localType]);
```
weili-govtech: instead of reverting via a useEffect, you can:
1. update localType to whatever value selected on `handleChange`
2. in `onClose`, revert to `selectedElementType`

tests/components/element-editor/basic-details/basic-details.spec.tsx
```diff
@@ -77,7 +77,7 @@ describe("BasicDetails", () => {
         });
 
         it("should replace element name, mandatory field's error message, and ID with values that correspond to the new element type", async () => {
```
weili-govtech: actually the fields got updated / dirtied because of the useEffect in `TestHelper` so it is not an effective test.

you can explore mocking the `showModal` and calling the callbacks from there

```
const mockHideModal = jest.fn();
const mockShowModal = jest.fn();
jest.mock("src/context-providers/display/modal-hook", () => {
	return {
		useModal: () => ({
			hideModal: mockHideModal,
			showModal: mockShowModal,
		}),
	};
});
// ...

it("should replace element name, mandatory field's error message, and ID with values that correspond to the new element type", async () => {
  // trigger reset confirmation modal
  expect(mockShowModal).toHaveBeenCalled();
  mockShowModal.mock.lastCall[1].onClickActionButton(); // <- this triggers the reset
});
```

weili-govtech:
create a new test suite for `TypeField` to test for the showing of the reset confirmation modal and the callbacks of `onClickActionButton` and `onClose`

weili-govtech (changes requested):

tests/components/element-editor/basic-details/basic-details.spec.tsx
adenteo: Thank you for this, did not know it was possible to trigger the actual callbacks.

src/util/test-helper.tsx
```diff
@@ -108,11 +108,17 @@ export namespace TestHelper {
...
+        useEffect(() => {
+            if (formContext?.currentValues) {
+                Object.keys(formContext.currentValues).forEach((key) => {
+                    methods.setValue(key, formContext.currentValues[key], {
+                        shouldDirty: true,
+                    });
+                });
+            }
+        }, [formContext?.currentValues, methods]);
+
```
weili-govtech: revert this change as it has no use and we dont want to cause any false positives

weili-govtech (changes requested):

weili-govtech (approved):

PR #129, opened 2024-09-06
CCUBE-1539

adenteo:
**Changes**
- Add customisation story to storybook.
- Add options prop to translator. 
- Add test case for hiding prefill.
- Add ternary operator to conditionally generate prefill object.
- Add test cases for translator parsing of prefill.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1539)


src/context-providers/config/types.ts
weili-govtech, line 1: sorry after some considerations, I think it will make more sense to put `prefill` under attributes. we will allow toggling of prefill for all elements only and not for individual elements, so with this, there should not be major changes to the logic.

weili-govtech (changes requested):
there is a need to hide the subtitle text for all preselected value fields because they mention the prefill value.

PR #134, opened 2024-09-12
CCUBE-1542

adenteo:
**Changes**
- Edit dropdownItems to be reusable component.
- Add support for radio button field. 
- Add test cases for radio items.

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1542)

PR #136, opened 2024-09-12
Fixes to radio field

weili-govtech:
**Changes**
- Continuation from https://github.com/LifeSG/web-form-builder/pull/134
- Contains misc fixes


tests/components/element-editor/basic-details/common/options.spec.tsx
```diff
@@ -19,61 +19,60 @@ describe("DropdownItems", () => {
...
 
-    describe("rendering the dropdown items", () => {
-        it("should render the dropdown items with 2 empty options if the element type is dropdown", async () => {
+    describe("rendering the items", () => {
+        it("should render the option items with 2 empty options if the element type is dropdown", async () => {
             renderComponent({
                 formContext: {
                     defaultValues: {
-                        dropdownItems: [
+                        optionItems: [
```
qroll: noob question but what determines the key of this field?

weili-govtech: upon looking at the tests again, I realised I shouldnt have updated the key, it should stay as `dropdownItems`. 

just a bit of background: this field is served as part of the basic details in the side panel (left panel). When admin clicks an already-added element in the main panel (right panel where a rough preview of the form is served). at any given time, we only serve one form, so upon serving in the side panel, we serve a new RHF instance and populate the context provider accordingly. 

in terms of the `dropdownItems` key, it is determined in either `DropdownBasicDetails` or `RadioButtonBasicDetails`, based on the `fieldName` passed.

qroll (approved):

PR #137, opened 2024-10-11
Add ability to hide pages panel

weili-govtech:
**Changes**
- Hide pages panel and toolbar if panels.pages.shouldShow = false
- Reduce width of side panel by ~1 column if toolbar is hidden
- Updated minimised side panel layout if toolbar is hidden


src/components/side-panel/side-panel-header.styles.tsx
```diff
@@ -14,14 +14,15 @@ interface IIconStyleProps {
 // =============================================================================
 // STYLING
 // =============================================================================
-export const Wrapper = styled.div`
+export const Wrapper = styled.div<{ $showDivider: boolean }>`
```
qroll: should introduce a style interface for this too

src/components/side-panel/side-panel.styles.tsx
```diff
@@ -21,7 +23,7 @@ export const Wrapper = styled.div<IWrapperStyleProps>`
     display: flex;
     flex-direction: column;
     height: calc(100% - ${(props) => props.$offset}rem);
-    width: 36.8rem;
+    width: ${({ $narrow }) => ($narrow ? "36.8rem" : "31.2rem")};
```
qroll: found it slightly confusing as the meaning of this variable is reversed? the element gets wider when `narrow` is true

qroll (changes requested):

src/components/side-panel/side-panel.styles.tsx
weili-govtech: oops realised I named in wrongly, I have reworked the values assigned

qroll (approved):

PR #143, opened 2025-08-20
Update packages

weili-govtech:
**Changes**
- Updated to use design system v3.0.0-alpha.20, frontend-engine v2.0.0-alpha.8
- Fixed npm vulnerabilities
- Fixed broken styles / test cases
- Updated jest to fix broken tests due to `:focus-visible` pseudo class, this caused the need to install a separate `@testing-library/jest-dom` and a manual import of the package


src/components/element-editor/basic-details/basic-details.styles.tsx
```diff
@@ -27,7 +27,7 @@ export const AccordionItem = styled(
 
     [data-testid="accordion-item-title"] {
         margin: ${({ $hideTopBorder }) =>
-            $hideTopBorder ? "1rem 0 1rem 1rem" : "0 0 0 1rem"};
+            $hideTopBorder ? "1rem 0 1rem 1rem" : "1rem 0 1rem 1rem"};
```
qroll: these are the same styles whether `hideTopBorder` is true or false, can remove the conditional check

src/components/element-editor/basic-details/basic-details.styles.tsx
```diff
@@ -36,8 +36,9 @@ export const AccordionItem = styled(
     }
 
     [data-testid="accordion-item-expand-collapse-button"] {
-        margin: ${({ $hideTopBorder }) =>
-            $hideTopBorder ? "0 0.1rem 0 1rem" : "0 0.1rem 0 0"};
+        margin-left: ${({ $hideTopBorder }) => ($hideTopBorder ? "1rem" : "0")};
```
qroll: becomes misaligned, margin left should remain 0 in this case, since the accordion-item-title already sets the required left offset

Screenshot 2025-08-28 at 1 09 19 PM

qroll (changes requested):

qroll (approved):

